PR #3, opened 2023-01-16
Label current db models as deprecated

joel-wright:
Pure renaming of imports from `app.db.` to `app.db.deprecated.` for existing db models


alembic/env.py
```diff
@@ -5,7 +5,7 @@
 from sqlalchemy import engine_from_config
 from sqlalchemy import pool
 
-from app.db.models import Base
+from app.db.deprecated.models import Base
```
joel-wright, line 8: @diversemix I'm not sure how this will interact with creating new db models & the multi-schema investigation, so worth looking at carefully! :grin:

diversemix (approved):
Looks good:
```
peter@uranus:~/work/navigator-backend (feature/mark-existing-db-models-as-deprecated %=)$ ag "from app.db.models" | grep -v deprecated
alembic/env.py:8:from app.db.models import Base
```

PR #4, opened 2023-01-17
Add data ingest script for bulk import

joel-wright:


scripts/data_ingest/main.py
```diff
@@ -0,0 +1,147 @@
...
+
+    :return: None
+    """
+    try:
+        data_ingest_response = post_data_ingest(ingest_csv_path=ingest_csv_path)
+    except Exception:
+        logger.exception("Calling the endpoint raised an unexpected exception.")
+        sys.exit(1)
+
+    if data_ingest_response.status_code == HTTPStatus.ACCEPTED:
```
diversemix, line 116: Do we want to dump out the returning `BulkImportValidatedResult` object?

diversemix (approved):
lgtm - one little comment

scripts/data_ingest/main.py
joel-wright: it does that in the info message below, or did you mean to stdout?

scripts/data_ingest/main.py
```diff
@@ -0,0 +1,147 @@
...
+
+import requests
+
+ADMIN_EMAIL_ENV = "SUPERUSER_EMAIL"
+ADMIN_PASSWORD_ENV = "SUPERUSER_PASSWORD"
+ADMIN_TOKEN_ENV = "SUPERUSER_TOKEN"
+BULK_IMPORT_ENDPOINT = "/api/v1/bulk-imports/cclw/law-policy"
+
+
+DEFAULT_LOGGING = {
```
THOR300, line 17: This logging configuration is different from the config we are using in the ingest stage which was an example / baseline of the logging format using json-logging just checking we're okay with that? I guess it doesn't really matter at all as we aren't querying these logs but just checking. 

See this below and here:
https://github.com/climatepolicyradar/navigator-data-ingest/blob/main/src/navigator_data_ingest/main.py#L30

scripts/data_ingest/main.py
```diff
@@ -0,0 +1,147 @@
...
+
+logging.config.dictConfig(DEFAULT_LOGGING)
+logger = logging.getLogger(__file__)
+
+
+def _log_response(response: requests.Response) -> None:
+    if response.status_code >= 400:
+        logger.error(
+            f"There was an error during a request to {response.url}. "
+            f"STATUS: {response.status_code}, BODY:{response.content!r}"
```
THOR300, line 52: extra: {props: {}} here?

scripts/data_ingest/main.py
```diff
@@ -0,0 +1,147 @@
...
+import logging.config
+import os
+import sys
+from http import HTTPStatus
+from pathlib import Path
+from pprint import pprint
+
+import requests
+
+ADMIN_EMAIL_ENV = "SUPERUSER_EMAIL"
```
THOR300, line 11: Integrating these via an .env file?

scripts/data_ingest/main.py
```diff
@@ -0,0 +1,147 @@
...
+        os.environ[ADMIN_TOKEN_ENV] = admin_user_token
+
+    return {
+        "Authorization": "Bearer {}".format(admin_user_token),
+        "Accept": "application/json",
+    }
+
+
+def get_request_url(endpoint: str) -> str:
+    api_host = os.getenv("API_HOST", "http://backend:8888").rstrip("/")
```
THOR300, line 90: Function comments for this and post_data_ingest functions?

THOR300 (approved):
LGTM just a couple minor queries.

joel-wright:
issues with multipart form data, marking as draft

scripts/data_ingest/main.py
joel-wright: okay, have checked & this is definitely now working as expected

scripts/data_ingest/main.py
joel-wright: this is just a temporary admin script for running locally until we have an admin interface, not something to be logging into CloudWatch, so I think this is okay

scripts/data_ingest/main.py
joel-wright: yeah, these are really just info logs for a local script, I think the default format is easier to read (these log lines are aimed at human admins)

scripts/data_ingest/main.py
joel-wright: I'll update the README with execution instructions instead...

scripts/data_ingest/main.py
joel-wright: I really should have written a get out of jail clause into those coding standards :joy:

joel-wright:
Fixed upload issues

PR #7, opened 2023-01-19
Feature/document updates and deletes

THOR300:
Facilitate the update and delete of documents.


app/core/util.py
```diff
@@ -73,3 +83,125 @@ def tree_table_to_json(
...
+) -> None:
+    """Delete the document objects in the relevant s3 buckets."""
+    if suffixes is None:
+        suffixes = [".json", ".npy"]
+
+    for prefix in prefixes:
+        for suffix in suffixes:
+            s3_path = S3Path(os.path.join("s3://", bucket, prefix, import_id + suffix))
+            if s3_path.exists():
+                s3_path.unlink()
```
diversemix, line 116: Might need to catch the `CloudPathIsADirectoryError` error

app/core/util.py
```diff
@@ -73,3 +83,125 @@ def tree_table_to_json(
...
+            csv_value=csv_document.create_request.category,
+            db_value=category_db_value,
+            updated=csv_document.create_request.category != category_db_value,
+        ),
+    )
+
+    _LOGGER.info(f"{update_results} for {csv_document.import_id}")
+    if any(
+        [
+            getattr(update_results, field.name).updated
```
diversemix, line 195: I can see why you did this - personally I'd prefer we get rid of `DocumentUpdateResults` and just define it as a `dict[str, UpdateResult]` that way you can do `update_results.any()` - but interesting :thinking:

app/core/util.py
```diff
@@ -73,3 +83,125 @@ def tree_table_to_json(
...
+    ):
+        update_doc_in_db(
+            updates=update_results, import_id=csv_document.import_id, db=db
+        )
+
+        delete_doc_in_s3(csv_document.import_id, PIPELINE_BUCKET, S3_PREFIXES)
+
+        return True
+
+    return False
```
diversemix, line 207: If we return `False` do we need to log a warning? Or info if the row is identical?

app/core/util.py
```diff
@@ -73,3 +83,125 @@ def tree_table_to_json(
...
+    updates: DocumentUpdateResults, import_id: str, db: Session
+) -> None:
+    """Update the document table in the database."""
+    with db.begin_nested():
+        for field in fields(DocumentUpdateResults):
+            if getattr(updates, field.name).updated:
+                db.query(Document).filter(Document.import_id == import_id).update(
+                    {field.name: getattr(updates, field.name).csv_value},
+                    synchronize_session="fetch",
+                )
```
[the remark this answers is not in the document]
THOR300: This is true for the update method. 

I can't seem to get the returning statement to work as it is not an attribute of the object. 

Update however does return an integer value for the no. of documents matched by the query and thus updated. 

Would it suffice to assert that the int == 1? 

Please see: 

"""
        :return: the count of rows matched as returned by the database's
         "row count" feature. 
"""

app/api/api_v1/schemas/document.py
```diff
@@ -177,6 +177,7 @@ class BulkImportValidatedResult(BaseModel):
 
     document_count: int
     document_added_count: int
+    document_updated_count: int
     document_skipped_count: int
```
THOR300, line 181: Suggesting changing the name "document_skipped_count". 

This is as we may see the following which could be confusing: 
document_count: 2 
document_added_count: 0 
document_updated_count: 1
document_skipped_count: 2

I would propose changing to the following or something similar:

document_count: 2 
document_ingest_count: 0 
document_existing_updated_count: 1
document_ingest_skipped_count: 2

diversemix: Even better `document_not_ingested_count` ? or even leave it off altogether if its `document_count` - `document_added_count`

app/core/config.py
```diff
@@ -65,3 +65,8 @@
     "OPENSEARCH_INDEX_ENCODER", "sentence-transformers/msmarco-distilbert-dot-v5"
 )
 OPENSEARCH_JIT_MAX_DOC_COUNT: int = int(os.getenv("OPENSEARCH_JIT_MAX_DOC_COUNT", "20"))
+
+PIPELINE_BUCKET: str = os.getenv("PIPELINE_BUCKET")
```
diversemix, line 69: don't you need to cast to `str()` to avoid the pyright warning?

app/core/util.py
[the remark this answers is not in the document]
diversemix: Yeah that would make sense - as we are only doing a field at a time

tests/routes/test_admin.py
```diff
@@ -512,3 +561,96 @@ def test_bulk_import_cclw_law_policy_preexisting_db_objects(
...
+    response = client.post(
+        "/api/v1/admin/bulk-imports/cclw/law-policy",
+        files=files,
+        headers=superuser_token_headers,
+    )
+    assert response.status_code == 202
+    response_json = response.json()
+    assert response_json["document_count"] == 2
+    assert response_json["document_updated_count"] == 0
+    assert response_json["document_skipped_count"] == 0
```
diversemix, line 625: Just realised we are returning two truths here: `document_skipped_count` and `len(document_skipped_ids)`

PR #8, opened 2023-01-20
Example of using the deprecated db model & new model

joel-wright:


joel-wright:
Example has served its purpose.

PR #10, opened 2023-01-23
BAK-1289 (1) Implement Document Schema

diversemix:
Cautiously starting to implement the new schema. Tested locally and setting up scripts to help migrate and test.

Specification: https://whimsical.com/database-schema-current-6s3aH2TfckVGpGJWhTgJ4j

Marked with green tick on document above:
image


THOR300 (approved):
LGTM - assuming the unique string will be what we now call import id?

app/db/models/document/physical_document.py
```diff
@@ -0,0 +1,45 @@
...
+class PhysicalDocument(Base):
+    """A physical document.
+
+    Representation of a document that exists in the real world.
+    Owned and updated only by the pipeline.
+    """
+
+    __tablename__ = "physical_document"
+
+    id = sa.Column(sa.Text, primary_key=True)
```
joel-wright, line 14: I'm not sure why we're setting the primary key to be text at this point; we have nothing sensible to set it to so why not use int?

app/db/models/document/physical_document.py
```diff
@@ -0,0 +1,45 @@
...
+    date = sa.Column(sa.DateTime, nullable=False)
+    format = sa.Column(sa.Text, nullable=True)
+
+
+class Language(Base):
+    """Note, moved from deprecated."""
+
+    __tablename__ = "language"
+
+    id = sa.Column(sa.SmallInteger, primary_key=True)
```
joel-wright, line 27: are we trying to save memory? I'm very nervous of ever using small integers...

app/db/models/document/physical_document.py
```diff
@@ -0,0 +1,45 @@
...
+    Representation of a document that exists in the real world.
+    Owned and updated only by the pipeline.
+    """
+
+    __tablename__ = "physical_document"
+
+    id = sa.Column(sa.Text, primary_key=True)
+    title = sa.Column(sa.Text, nullable=False)
+    md5_sum = sa.Column(sa.Text, nullable=True)
+    source_url = sa.Column(sa.Text, nullable=True)
```
joel-wright, line 17: I know I set this to `varchar` in the table, but maybe we can use a URL type here?

joel-wright (commented):
A few comments about the physical doc field types...

app/db/models/document/physical_document.py
diversemix: ... this was blindly copied from the deprecated module - we can migrate it I suppose?

app/db/models/document/physical_document.py
joel-wright: Is that not pushing law/policy domain information into the application agnostic document schema layer?

app/db/models/document/physical_document.py
joel-wright: ugh, I knew I'd seen this before! Might be worth taking the time to migrate now...

app/db/models/document/physical_document.py
joel-wright: also... we have a single "application schema" at the moment, and while it's true that the family-document <-> physical-document is 1:1 for that app, can we guarantee that constraint holds for all future apps?

app/db/models/document/physical_document.py
joel-wright: ok, this seems to require a custom extension, so scratch that idea

app/db/models/document/physical_document.py
joel-wright: `int2` really scares me

app/db/models/document/physical_document.py
diversemix: 1) It's not really pushing law/policy into this layer - its allowing a upper layer to reference the document schema in a way that it understands - thinking of document id as an abstract identifier that is actually defined in the upper layer... if that makes sense

2) This also re-inforces the above idea - in that even in other apps where the reln is not 1:1 ... the definition of the id needs to really be specified at the upper-scheama level

diversemix: FTR - conversation concluded to use an ID

app/db/models/document/physical_document.py
```diff
@@ -0,0 +1,44 @@
...
+    """A document's languages."""
+
+    __tablename__ = "physical_document_language"
+
+    id = sa.Column(sa.Integer, primary_key=True)
+    language_id = sa.Column(sa.ForeignKey(Language.id), nullable=False)
+    document_id = sa.Column(
+        sa.ForeignKey(PhysicalDocument.id, ondelete="CASCADE"),
+        nullable=False,
+    )
```
joel-wright, line 44: When we spoke earlier I thought we agreed to remove the int id on these join tables...?

```suggestion
    language_id = sa.Column(sa.ForeignKey(Language.id), nullable=False)
    document_id = sa.Column(
        sa.ForeignKey(PhysicalDocument.id, ondelete="CASCADE"),
        nullable=False,
    )
    
    PrimaryKeyConstraint(language_id, document_id)
```

joel-wright: We can always do this in the follow-up

joel-wright (approved):

app/db/models/document/physical_document.py
diversemix: FTR - this was done at the point it [was raised](https://github.com/climatepolicyradar/navigator-backend/blob/3855ca5939fa0d5a7741e01ad92c0dc0fb4c7301/app/db/models/document/physical_document.py#L34-L45) ... but closing this off now and merging 1&2 as agreed

diversemix:
Closing as merged (1) and (2)

PR #11, opened 2023-01-23
BAK 1289 (2) Implement Application Schema

diversemix:
See https://whimsical.com/database-schema-current-6s3aH2TfckVGpGJWhTgJ4j

image


app/db/models/app/users.py
```diff
@@ -0,0 +1,36 @@
...
+class AdminOrganisation(Base):
+    """Link table for admin and organisation.  """
+
+    __tablename__ = "admin_organisation"
+
+    admin_id = sa.Column(sa.Integer, sa.ForeignKey(Admin.id), nullable=False)
+    organisation_id = sa.Column(sa.Integer, sa.ForeignKey(Organisation.id), nullable=False)
+    job_title = sa.Column(sa.String)
+
+    PrimaryKeyConstraint(admin_id, organisation_id)
```
joel-wright, line 36: A few questions here:
 - Do we want to give this PK a name, or are we happy yo just use the default?
 - Is this how we want to specify PK constraints for join tables, because it doesn't seem to be consistent with e.g. `PhysicalDocumentLanguage` (I'm not saying it doesn't make sense, it really does, but consistency is good too :)
 - Does defining the PK this way introduce any difficulties?

joel-wright (commented):
Just a single multi-faceted question :sweat_smile:

app/db/models/app/users.py
diversemix: FTR - talked about this and we're going to make it consistent - removing the id and using the two columns in a link table

joel-wright (approved):
lgtm, just the pk changes in branch (1)

PR #15, opened 2023-01-25
Ensure the app folder is copied correctly

diversemix:


THOR300 (approved):
Tested locally and looks good to me: 
building the image and running "docker run" without mounting the volumes fires off the correct command!

PR #17, opened 2023-01-25
Fixing dockerfile.

THOR300:
App runner deployment is still failing as alembic directory is not copied correctly.


diversemix (approved):
great

PR #20, opened 2023-01-27
Add script for ingesting into new schema

diversemix:
- Implement Document Schema
- tidy
- Add migration for new tables
- Implement new app schema
- wip
- wip
- fixup and add migration
- fixup tests
- end of day checkin
- re-generate migration
- Add the metadata tables
- Use JSONB columns from postgres not JSON type
- end-of-day
- end-of-day
- Fix id's to be text
- update primary key types
- trigger GA
- wip
- update primary keys
- fix breakage after documenttype change
- end-of-week


diversemix:
opened by mistake

PR #22, opened 2023-02-02
BAK-1291 Create new ingest script for CSV

diversemix:
Tested locally and as far as I can see it works. There are rows created as expected in the database, still more work to do on the script around the parsing and validation of certain fields.


scripts/ingest_dfc/dfc_csv_reader.py
```diff
@@ -0,0 +1,134 @@
...
+        assert set(REQUIRED_COLUMNS).issubset(set(reader.fieldnames))
+        row_count = 0
+        errors = False
+
+        for row in reader:
+            row_count += 1
+            row_object = Row(row_count, row)
+            if row_count >= row_start:
+                if not process(row_object):
+                    break
```
THOR300, line 131: Assuming this is to be an errors = True as I can't see where errors = False would ever be switched to True here?

diversemix: Early days for the script - I'm going to consider errors when its working across more than a couple of items

scripts/ingest_dfc/dfc/processor.py
```diff
@@ -0,0 +1,96 @@
...
+            f"Found {num_new_documents} new documents and {num_old_documents} old "
+            "documents"
+        )
+        return True  # num_new_documents == 0 and num_old_documents > 0
+
+    def process(row: DfcRow) -> None:
+        """Processes the row into the db."""
+        sys.stdout.write(f"Processing row: {row.row_number}: ")
+
+        # Beginning a transaction here would create this issue:
```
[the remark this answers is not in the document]
joel-wright: kk

joel-wright (approved):
This has certainly been around enough cycles of testing & review that more are unlikely to be much help :grin:

PR #28, opened 2023-02-10
Fix getting the source_url

diversemix:


scripts/ingest_dfc/utils.py
```diff
@@ -153,6 +153,17 @@ def _key(key: str) -> str:
...
 
 
+    def get_first_url(self) -> str:
+        """Gets the first URL from the 'documents' attribute.
+        
+        TODO: This could/should be written with more validation.
+        """
+        first_doc = self.documents.split(';')[0]
+        first_url = first_doc.split('|')[0]
+        return first_url
```
joel-wright, line 163: Is it worth warning/erroring if we find more than one document?

diversemix: Yeah - good point - should I raise ?

joel-wright (approved):

PR #29, opened 2023-02-13
BAK-1305 Create and populate metadata_taxonomy

diversemix:
The first of several PRs to populate and use the taxonomy on ingestion.

### DIFF
--- a/app/data_migrations/populate_taxonomy.py
+++ b/app/data_migrations/populate_taxonomy.py
@@ -1,0 +1,110 @@
+import json
+from typing import Sequence
+
+from sqlalchemy.orm import Session
+
+from app.db.models.law_policy.metadata import MetadataTaxonomy
+
+from .utils import has_rows
+
+"""At the moment taxonomy is kept simple, and only supports string validation for enums
+
+For Example: 
+
+{
+    "topic": {
+       allowed_values: [],
+       allow_blanks: false,
+    },
+    ...
+}
+
+"""
+
+TAXONOMY_DATA = [
+    {
+        "key": "topic",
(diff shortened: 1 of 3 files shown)


app/data_migrations/populate_taxonomy.py
```diff
@@ -0,0 +1,110 @@
...
+def populate_taxonomy(db: Session) -> None:
+    """Populates the taxonomy from the data."""
+
+    if has_rows(db, MetadataTaxonomy):
+        return
+
+    db.add(MetadataTaxonomy(
+        name="default", 
+        description="CCLW loaded values", 
+        valid_metadata=get_default_taxonomy()))
```
joel-wright, line 110: we could load the initial data from the json files - doing it from the existing db rows will be a pain when we remove the old metadata tables...

joel-wright: okay, may have misread it - brb

app/data_migrations/populate_taxonomy.py
```diff
@@ -0,0 +1,110 @@
...
+    if "." not in dotted_key:
+        return obj[dotted_key]
+    keys = dotted_key.split(".", 1)
+    return dot_dref(obj[keys[0]], keys[1])
+    
+
+def load_metadata_type(filename: str, key_path: str) -> Sequence[str]:
+    with open(filename) as file:
+        data = json.load(file)
+    return [dot_dref(obj, key_path) for obj in data]
```
diversemix, line 85: @joel-wright This is the function that loads the metadata from the data files :smile:

joel-wright (commented):
LGTM

joel-wright (approved):

PR #30, opened 2023-02-13
Schema updates

diversemix:
- Changes slug to ref document import_id (not the physical document )
- Adds family status
- Makes family and document status non-nullable


joel-wright (approved):
LGTM

PR #34, opened 2023-02-15
Update pre-commit checks

joel-wright:
Updates:
 - new pyright version
 - switch from flake8/pydocstyle to ruff
 - fixes newly flagged issues


diversemix (approved):
Gold - solid gold :1st_place_medal:

PR #38, opened 2023-02-20
Add PR template

joel-wright:


.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
+# Description
```
diversemix, line 1: Do we want to add a reference to linear ?

joel-wright: let's look at this as we go forward...

PR #40, opened 2023-02-23
Feature/events

joel-wright:
# Description

This PR introduces the ability to import CCLW Event data & associate it with `Familiy`s. It also marks imported Event data as "duplicated" if the import process could not determine unambiguously which `Family` should be associated with an `Event`. We will make use of these tags in a future data cleaning exercise.

Also includes some minor drive-by fixes to our new database structure.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

WIP: no tests yet!

## Reviewer Checklist

- [ ] No FIXMEs remain


app/api/api_v1/routers/admin.py
```diff
@@ -349,48 +422,108 @@ def start_ingest(db: Session, s3_client: S3Client, file_contents: str):
...
             }
         },
     )
 
     # TODO: Add some way the caller can monitor processing pipeline...
-
-    return csv_s3_location
+    return BulkImportResult(
+        import_s3_prefix=s3_prefix,
+        detail=None,  # TODO: add detail?
```
diversemix, line 518: We could put the detail from the validation result on here

app/db/models/law_policy/family.py
```diff
@@ -9,11 +12,24 @@
...
+class _BaseModelEnum(str, enum.Enum):
     """Family categories as understood in the context of law/policy."""
 
-    EXECUTIVE = "EXECUTIVE"
-    LEGISLATIVE = "LEGISLATIVE"
+    @classmethod
+    def _missing_(cls, value):
+        if isinstance(value, str) and not str.istitle(value):
+            return cls(value.title())
+        raise ValueError(f"{value} is not a valid FamilyCategory")
```
diversemix, line 22: This error references `FamilyCategory` in `_BaseModelEnum`

joel-wright: sorted

app/api/api_v1/routers/admin.py
joel-wright: ah, right, have looked at this. I agree we could put this detail in, but probably something for a follow up because `detail` here has a very specific type & expected content.

PR #84, opened 2023-04-05
Validate for consistent families and collections

diversemix:
# Description

- Introduces `MultiplyDefinedEntities` into the ingest context
- These are used on validation
- When inconsistent errors are added to validation output
- Tests added

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/core/ingestion/utils.py
+++ b/app/core/ingestion/utils.py
@@ -156,12 +156,3 @@ def update_if_changed(updated: dict, updated_key: str, source: Any, dest: Any):
         setattr(dest, updated_key, source)
         updated[updated_key] = source
     return updated
-
-
-class IngestOperation(str, enum.Enum):
-    """Type of operation the Ingest is going to perform for a row,"""
-
-    # MIGRATE = "Migrate"
-    CREATE = "Create"
-    UPDATE = "Update"
-    # DELETE = "Delete" - TODO: Still to be implemented
--- a/app/core/ingestion/utils.py
+++ b/app/core/ingestion/utils.py
@@ -151,6 +151,13 @@ def get_result_counts(results: list[Result]) -> tuple[int, int, int]:
     return rows, fails, resolved
 
 
+def update_if_changed(updated: dict, updated_key: str, source: Any, dest: Any):
+    if dest != source:
+        dest = source
+        updated[updated_key] = source
+    return updated
+
+
 class IngestOperation(str, enum.Enum):
     """Type of operation the Ingest is going to perform for a row,"""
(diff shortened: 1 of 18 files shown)


app/core/ingestion/utils.py
```diff
@@ -128,12 +128,29 @@ class Result:
...
+    collections: dict[str, NameSummary]  # id -> name, description
+
+
 @dataclass
 class IngestContext:
     """Context used when processing."""
 
     org_id: int
     results: list[Result]
+    mde: MultiplyDefinedEntities
```
joel-wright, line 148: `mde` really isn't what I'd call a "well-known acronym", so maybe we can spell it out? Also, I'm not sure that `MultiplyDefinedEntities` really expresses what this thing is for. Would something like `ValidationLookup` no express more what this class is for?

app/core/ingestion/validator.py
```diff
@@ -55,10 +59,59 @@ def validate_document_row(
...
+    # Check metadata
     result, _ = build_metadata(taxonomy, row)
     if result.type != ResultType.OK:
         errors.append(result)
 
+    on_row = f"on row {row.row_number}"
+    # Check family
+    family_id = row.cpr_family_id
+
+    if family_id in context.mde.families.keys():
```
joel-wright, line 71: Not only is this call to `keys()` redundant, it turns an O(1) lookup into an O(n) lookup... :grin:

app/core/ingestion/validator.py
```diff
@@ -55,10 +59,59 @@ def validate_document_row(
...
+                    f"Family {family_id} has differing summary {on_row}",
+                )
+            )
+    else:
+        context.mde.families[family_id] = (row.family_name, row.family_summary)
+
+    # Check collection
+    collection_id = row.cpr_collection_id
+
+    if collection_id in context.mde.collections.keys():
```
joel-wright, line 93: Similar complexity issue to above.

joel-wright (approved):
A few comments, but it's definitely going to make our lives better!

app/core/ingestion/utils.py
```diff
@@ -128,12 +128,29 @@ class Result:
     details: str = ""
 
 
+NameSummary = tuple[str, str]
```
joel-wright, line 131: It may be worth updating this to be a "proper" dataclass, because we'll also need to support deletion marks soon...

PR #98, opened 2023-04-26
Filter climate-laws.org source URLs that will no longer work from responses

joel-wright:
# Description

Do not return climate-laws.org source URLs from API responses because they will no longer work after domain switch

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Response models have been tested to confirm that relevant URLs are removed

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/tests/unit/app/schemas/test_schemas.py
+++ b/tests/unit/app/schemas/test_schemas.py
@@ -1,0 +1,100 @@
+import pytest
+
+from app.api.api_v1.schemas.document import FamilyDocumentResponse
+from app.api.api_v1.schemas.search import SearchResponseFamilyDocument
+
+CLIMATE_LAWS_DOMAIN_PATHS = [
+    "climate-laws.org",
+    "climate-laws.org/path",
+    "climate-laws.org/path/multiple/elements/",
+    "anything.climate-laws.org/",
+    "anything.climate-laws.org/path/",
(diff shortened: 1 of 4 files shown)


tests/unit/app/schemas/test_schemas.py
```diff
@@ -0,0 +1,100 @@
...
+    "sub.not-climate-laws.org",
+    "example.com/path",
+    "not-climate-laws.org/path/multiple/elements/",
+]
+SCHEMES = ["http", "https"]
+
+
+@pytest.mark.parametrize("source_domain_path", CLIMATE_LAWS_DOMAIN_PATHS)
+@pytest.mark.parametrize("scheme", SCHEMES)
+def test_climate_laws_source_url_filtered_from_search(source_domain_path, scheme):
```
diversemix, line 27: Do we also want test(s) for browse too - just for completeness :grin:

diversemix (approved):
lgtm - a "take it or leave it" comment - assuming browse uses `DocumentOverviewResponse`

tests/unit/app/schemas/test_schemas.py
joel-wright: We do need better tests for this filtering, I will raise a ticket

PR #107, opened 2023-05-06
Resolve FIXMEs

joel-wright:
# Description

Resolve FIXMEs

## Type of change

- [x] Cleanup
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Tests unchanged or extended where required.

## Reviewer Checklist

- [x] No FIXMEs remain

### DIFF
--- a/app/core/browse.py
+++ b/app/core/browse.py
@@ -59,10 +59,11 @@ def to_search_response_family(
         family_category=str(family.family_category),
         family_date=family_published_date,
         family_last_updated_date=family_last_updated_date,
-        family_source=cast(str, organisation.name),  # FIXME: check links to source
+        family_source=cast(str, organisation.name),
         family_geography=cast(str, geography.value),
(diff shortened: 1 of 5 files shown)


app/core/browse.py
```diff
@@ -59,13 +59,14 @@ def to_search_response_family(
         family_category=str(family.family_category),
         family_date=family_published_date,
         family_last_updated_date=family_last_updated_date,
-        family_source=cast(str, organisation.name),  # FIXME: check links to source
```
THOR300, line 62: I don't entirely understand what this fixme means and thus can't ascertain whether it's now met?

THOR300 (approved):
LGTM

app/core/browse.py
joel-wright: This was just leftover from partial implementation of the browsing functionality. We're now getting the source name directly from the family link in the database, so it can no longer be wrong.

PR #131, opened 2023-05-31
Feature/update document language via endpoint

THOR300:
# Description

[Linear Ticket](https://linear.app/climate-policy-radar/issue/SOF-10/language-detection-from-the-pipeline-back-to-the-backend-so-that)

Please include:
- a summary of the changes

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/api/api_v1/routers/admin.py
```diff
@@ -71,10 +71,39 @@ async def update_document(
...
     ).rowcount
 
+    # Update the languages
+    if meta_data.languages is not None:
+        _LOGGER.info(
+            "Adding meta_data object languages to the database.",
+            extra={"props": {'meta_data_languages': meta_data.languages}}
+        )
+
+        physical_document_languages = (
```
diversemix, line 85: Why can't lines 85-96 be done with a single `db.delete` statement?

app/api/api_v1/routers/admin.py
```diff
@@ -71,10 +71,39 @@ async def update_document(
...
+            db.query(PhysicalDocumentLanguage)
+            .filter(
+                PhysicalDocumentLanguage.document_id == physical_document.id,
+            )
+            .all()
+        )
+
+        if physical_document_languages is not None:
+            for physical_document_language in physical_document_languages:
+                db.delete(physical_document_language)
```
diversemix, line 95: Might we not be removing rows that we are then going to re-create?

tests/routes/test_document_families.py
```diff
@@ -323,6 +325,65 @@ def test_update_document__works_on_import_id(
...
+    assert doc.content_type == "updated/content_type"
+    assert doc.cdn_object == "folder/file"
+
+    languages = (
+        test_db.query(PhysicalDocumentLanguage)
+        .filter(PhysicalDocumentLanguage.document_id == doc.id)
+        .all()
+    )
+    assert len(languages) == 2
+    assert [l.language_id for l in languages] == [2, 1]
```
diversemix, line 384: This could be a little fragile as it depends on the order - prefer using `set`

app/api/api_v1/routers/admin.py
```diff
@@ -71,10 +71,29 @@ async def update_document(
...
+        db.query(PhysicalDocumentLanguage).filter(PhysicalDocumentLanguage.document_id == physical_document.id).delete()
+        db.flush()
+
+        for language in meta_data.languages:
+            lang = db.query(Language).filter(Language.language_code == language).one_or_none()
+            if lang is not None:
+                physical_document_language = PhysicalDocumentLanguage(
+                    language_id=lang.id, document_id=physical_document.id
+                )
+                db.add(physical_document_language)
```
joel-wright, line 94: This will throw errors if a language association already exists

tests/routes/test_document_families.py
```diff
@@ -323,6 +325,65 @@ def test_update_document__works_on_import_id(
...
 
 
+@pytest.mark.parametrize(
+    "import_id",
+    [
+        "CCLW.executive.1.2",
+        "UNFCCC.non-party.2.2",
+    ],
+)
+def test_update_document__works_on_new_language(
```
joel-wright, line 335: I think we also need tests for setting languages on a document that already has languages set, both where the updates overlap with the existing ones, and where they don't. We need to confirm that we end up with the combination of existing & updates each time and that no errors are thrown.

joel-wright (commented):
It's a good start, but we need to cope with the case where a document already has a language (or even languages) set.

app/api/api_v1/routers/admin.py
THOR300: I'm not sure I follow why? 
On line 85 we remove all PhysicalDocumentLanguage associations from the database and then re-populate the associations from the languages declared in the request object?

tests/routes/test_document_families.py
THOR300: Surely the initial documents that are loaded into the database via the following code on line 343 would load have the language Algeria already? 

python```
setup_with_multiple_docs(
        test_db, mocker, doc_data=TWO_DFC_ROW_DIFFERENT_ORG, event_data=TWO_EVENT_ROWS
)
```

https://github.com/climatepolicyradar/navigator-backend/blob/feature/update-document-language-via-endpoint/tests/routes/document_helpers.py#L26

My understanding is that we go: 
1. Algeria 
2. Delete data
3. Load English and French

app/api/api_v1/routers/admin.py
joel-wright: I missed the deletion... but I can't say I'm super comfortable with the approach. If we delete all the language associations, then fail during updates we lose information. Is it possible to calculate the difference, and therefore the ones we need to add?

joel-wright (commented):
Generally looks good, but would prefer the language update to be calculated as an addition rather than deleting all existing languages for now.

joel-wright (commented):
This looks good - we'll need to keep an eye on the impact of lots of database queries, but for now with small numbers of languages I think we should be okay. I just have one request for another quick test that adds multiple languages (to make sure that the multiple add queries don't clash with things like ID generation).

joel-wright (approved):

PR #143, opened 2023-06-19
Minor CSV download content fix

joel-wright:
# Description

Previous fix did not correctly identify when the content of a document matched a given search term. This is no longer the case.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Existing tests updated after testing that regressions would be caught.

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/tests/routes/test_search.py
+++ b/tests/routes/test_search.py
@@ -1259,15 +1259,16 @@ def test_csv_content(
                 document = validation_data[family_name]["documents"][doc_title]
                 assert document["document_title"] == row["Document Title"]
                 assert row["Document URL"].endswith(document["document_slug"])
+                # Check that if the content type is pdf, we include a CDN URL for
+                # the document, otherwise we send the document source URL.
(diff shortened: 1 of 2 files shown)


tests/routes/test_search.py
```diff
@@ -1264,7 +1265,8 @@ def test_csv_content(
                     )
                 else:
                     assert (
-                        document["document_source_url"] == row["Document Content URL"]
+                        document["document_source_url"]
```
diversemix, line 1268: Is there a way to make this more explicit about what we are testing for. The code is fine but I'm not sure what we are asserting for - the complexity of the statement here is a little obtuse for me - or maybe we should add a comment?

diversemix (approved):
On faith

tests/routes/test_search.py
diversemix: Perfect :ok: thanks

PR #155, opened 2023-07-10
Making updates to handle different iso codes.

THOR300:
# Description

The current backend code assumes that language codes send to the update document endpoint will be if the for iso-639-3. This is a three letter representation of the language. 

The pipeline identifies languages as iso-639-1 which is a two letter representation. 

Thus, the endpoint is updated to handle the differing language codes. A unit test is also added to check functionality. 

NOTE: 
Should we update the languages table to enforce part1_codes being unique and len(2) as the functionality will rely on that an in practice the two letter codes should be unique and two letters (the same of language codes and three letters)? 
Should we also at least log when we pass over a language update due to it not existing in the database? 

[Linear Ticket](https://linear.app/climate-policy-radar/issue/PDCT-200/bugfixhandle-iso-639-1-language-codes)

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Unit test in the PR 

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/admin.py
+++ b/app/api/api_v1/routers/admin.py
@@ -83,7 +83,12 @@ async def update_document(
     if meta_data.languages is not None:
         _LOGGER.info(
             "Adding meta_data object languages to the database.",
-            extra={"props": {"meta_data_languages": meta_data.languages}},
+            extra={
+                "props": {
+                    "meta_data_languages": meta_data.languages,
+                    "import_id_or_slug": import_id_or_slug,
+                }
+            },
         )
 
         physical_document_languages = (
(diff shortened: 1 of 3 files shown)


app/api/api_v1/routers/admin.py
```diff
@@ -97,11 +97,23 @@ async def update_document(
...
+                    db.query(Language)
+                    .filter(Language.part1_code == language)
+                    .one_or_none()
+                )
+            elif len(language) == 3:
+                lang = (
+                    db.query(Language)
+                    .filter(Language.language_code == language)
+                    .one_or_none()
+                )
```
joel-wright, line 113: I thought we were falling back from language_code to part2_code?

THOR300: Logging a warning if the language is None. This is the only scenario we need to be aware of. If we retrieve a language and it doesn't exist then we will add it.

tests/routes/test_document_families.py
```diff
@@ -516,6 +516,183 @@ def test_update_document__works_on_new_language(
...
+    [
+        "CCLW.executive.1.2",
+        "UNFCCC.non-party.2.2",
+    ],
+)
+def test_update_document__works_on_new_iso_639_1_language(
+    client: TestClient,
+    superuser_token_headers: dict[str, str],
+    test_db: Session,
+    mocker: Callable[..., Generator[MockerFixture, None, None]],
```
joel-wright, line 530: :joy: nice!

joel-wright (approved):
Looking good :)

PR #161, opened 2023-08-02
PDCT 251/improve lang

diversemix:
# Description

- Adds new flags to the PhysicalDocumentLanguage
- Adds db migrations
- Updates tests for pipeline to use the `source` = `MODEL`
- Updates ingestion to use `source` = `USER`
- Renames `_BaseModelEnum`  to `_BaseModelEnum` moved into new file to avoid circular imports

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Test Result

The following was tested locally:
- start from blank database creates correct schema
- start from existing staging data - also creates correct schema with the defaults

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/tests/routes/test_document_families.py
+++ b/tests/routes/test_document_families.py
@@ -623,8 +623,8 @@ def test_update_document__works_on_new_iso_639_1_language(
     assert json_object["md5_sum"] == "c184214e-4870-48e0-adab-3e064b1b0e76"
     assert json_object["content_type"] == "updated/content_type"
     assert json_object["cdn_object"] == "folder/file"
-    expected_languages = ["ell", "bod"]
-    assert json_object["languages"] == expected_languages
+    expected_languages = set(["ell", "bod"])
+    assert set(json_object["languages"]) == expected_languages
 
     # Now Check the db
     doc = (


app/core/ingestion/family.py
```diff
@@ -183,6 +186,9 @@ def none_if_empty(data: str) -> Optional[str]:
                 params.document_title,
                 family_document.physical_document,
             )
+            update_physical_document_languages(
+                db, params.language, result, family_document.physical_document
+            )
```
joel-wright, line 191: Is deleting all existing languages on every ingest not a little heavy handed? I had wondered if we could label all existing languages as not visible with a one-off script, then this script method would only need to add a missing language or mark an existing one as "user + visible" if necessary... 

Also, this makes it impossible for the pipeline to add document languages unless we run a complete update each time (currently we only send updates).

app/db/crud/document.py
```diff
@@ -102,21 +107,27 @@ def get_family_document_and_context(
...
+) -> Sequence[str]:
+    result = (
+        db.query(PhysicalDocumentLanguage.visible, Language.language_code)
+        .join(Language, Language.id == PhysicalDocumentLanguage.language_id)
+        .filter(PhysicalDocumentLanguage.document_id == physical_document.id)
+        .all()
+    )
+    if result is None or len(result) == 0:
+        return []
+    return [cast(str, code) for visible, code in result if visible]
```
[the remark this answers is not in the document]
diversemix: I did think about this - but let's chat about it, as its not a standard "relationship" as it requires a filter - so ends up being a property/function anyhow

app/core/ingestion/family.py
diversemix: I don't believe the code does delete all existing languages on every ingest - perhaps you can point out where you think this is occurring?

app/core/ingestion/physical_document.py
```diff
@@ -32,21 +36,53 @@ def create_physical_document_from_params(
...
-            result["physical_document_language"] = phys_doc_languages
+        doc_languages = result.get("language", [])
+        doc_languages.append(to_dict(lang))
+        result["language"] = doc_languages
 
-    return physical_document
+        physical_document_language = PhysicalDocumentLanguage(
+            language_id=lang.id,
+            document_id=physical_document.id,
+            source=LanguageSource.USER,
```
THOR300, line 80: I'm just wondering whether this function would be utilised by the db writer when calling back to the backend with identified languages and thus whether defaulting to a USER defined language is correct.

alembic/versions/0018_add_more_language_fields.py
```diff
@@ -0,0 +1,54 @@
...
+    #
+    # Code taken from: https://stackoverflow.com/a/65173731
+    language_source = postgresql.ENUM(LanguageSource, name="languagesource")
+    language_source.create(op.get_bind(), checkfirst=True)
+
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('physical_document_language', sa.Column(
+        'source',
+        language_source,
+        server_default=(LanguageSource.MODEL.upper()),
```
THOR300, line 33: Just to confirm for all the physical documents we currently have in the database we would be assigning a default value of MODEL defined language?

app/core/ingestion/physical_document.py
diversemix: Nope this is ingestion only - i.e. via CSV ... i.e. via a user :smile:

alembic/versions/0018_add_more_language_fields.py
diversemix: Yup - and also `visible` = False

joel-wright (approved):

PR #162, opened 2023-08-07
Fix setting language link to USER/visible during ingest

joel-wright:
# Description

This PR updates the ingest code to switch language links supplied in an ingested CSV to "supplied by the user & should be visible" when appropriate.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Additional test to validate the change.

## Reviewer Checklist

- [ ] No FIXMEs remain


pyproject.toml
```diff
@@ -45,6 +45,7 @@ pyright = "^1.1.294"
 pytest-asyncio = "^0.18.3"
 ruff = "^0.0.246"
 pytest-alembic = "^0.10.5"
+types-SQLAlchemy = "^1.4.31"
```
THOR300, line 48: Where is this used?

THOR300, line 48: Where is this used?

THOR300 (approved):
LGTM just thinking whether we need to add a test that mimics the db writer update and then assert that the document language that's added is not visible.

joel-wright:
> LGTM just thinking whether we need to add a test that mimics the db writer update and then assert that the document language that's added is not visible.

I think this is already covered by the `test_document_families.py` tests - there are tests there that assert that the languages are added with a source of `LanguageSource.MODEL`... is that what you meant?

PR #168, opened 2023-10-02
PDCT 395/only add counters

diversemix:
# Description
- Adds a counter table
- Ability to generate a unique import_id and increment atomically
- Adds a migration to add counter for CCLW and UNFCCC
- Adds tests

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/db/models/law_policy/family.py
```diff
@@ -17,7 +17,7 @@ class FamilyCategory(BaseModelEnum):
 
     EXECUTIVE = "Executive"
     LEGISLATIVE = "Legislative"
-    UNFCCC = "UNFCCC"
+    UNFCCC = ORGANISATION_UNFCCC
```
diversemix, line 20: Hmmmm :thinking: ... maybe this should not be used here.

alembic/versions/0019_add_entity_counters.py
```diff
@@ -0,0 +1,40 @@
...
+Revision ID: 0019
+Revises: 0018
+Create Date: 2023-10-02 11:32:43.825217
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.orm import Session
+
+from app.db.models.app.counters import ORGANISATION_CCLW, ORGANISATION_UNFCCC, EntityCounter
```
joel-wright, line 12: where are these imports used?

app/db/models/law_policy/family.py
joel-wright: yeah, probably not

app/db/models/app/counters.py
```diff
@@ -0,0 +1,95 @@
...
+            db = object_session(self)
+            cmd = self._get_and_increment.bindparams(id=self.id)
+            value = db.execute(cmd).scalar()
+            db.commit()
+            return value
+        except:
+            _LOGGER.exception(f"When generating counter for {self.prefix}")
+            raise
+
+    def get_import_id(self, entity: CountedEntity, n: int = 0) -> str:
```
joel-wright, line 84: relying on the rest of the code to generate the n-value seems risky, especially because clashes won't be detected until a db transaction fails...

joel-wright (commented):
I can't see where these counters are being used or tested... what am I missing?

app/db/models/app/counters.py
diversemix: Good thought :thinking:

alembic/versions/0019_add_entity_counters.py
diversemix: Ha I removed the code - let me fix that up - thanks :+1:

app/db/models/app/counters.py
diversemix: ok - thinking about this - I believe this code is fine - as this generates a new import_id - I believe what we need is another function that generates the next valid n-value.

joel-wright: I think I see what you're saying here & agree with you when it comes to adding families. Perhaps we need to either move away from the structure of import IDs we have at the moment for things like documents (which currently share a 3rd quad with the family), and just always assign 0 to the 4th quad (and only to match the current import_id checking). That way anything generated by the new code is always "<source>.<type>.<guaranteed_safe_count>.0" and any implied meaning we had in previous import IDs is no longer required or expected.

I guess what I'm saying here is just "the n doesn't need to be parameterised, just always use 0" :)

joel-wright (approved):
LGTM

PR #179, opened 2023-11-07
Bump data-access & tweak result hits value for Vespa search

joel-wright:
# Description

Bumps data access version to fix an internal server error on 0 results, and tweak the value of hits reported to indicate the number of families returned by a response.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Existing tests do not need to be updated - the error case was entirely contained in data-access library which now has updated tests.

## Reviewer Checklist

- [ ] No FIXMEs remain


app/core/search.py
```diff
@@ -1227,7 +1227,7 @@ def process_vespa_search_response(
 ) -> SearchResponse:
     """Process a Vespa search response into a F/E search response"""
     return SearchResponse(
-        hits=vespa_search_response.total_hits,
+        hits=len(vespa_search_response.families),
```
olaughter, line 1230: Are we still happy referring to this as 'hits'?

olaughter (approved):
lgtm, one question below

app/core/search.py
joel-wright: this actually matches the current Opensearch definition of hits, and @harrisonpim is raising a ticket in data science to look into making "hits" a more accurate measure of the number of families that would match rather than cutting it of at the query limit

PR #180, opened 2023-11-09
Add hybrid expression for family status in db model.

katybaulch:
Added hybrid expression for family status hybrid property in backend DB model for Family entity in keeping with [this PR](https://github.com/climatepolicyradar/navigator-admin-backend/pull/50/files).

### DIFF
--- a/app/db/models/law_policy/family.py
+++ b/app/db/models/law_policy/family.py
@@ -65,7 +65,7 @@ class Family(Base):
     )
 
     @hybrid_property
-    def family_status(self) -> FamilyStatus:
+    def family_status(self) -> FamilyStatus:  # type: ignore
         """Calculates the family status given its documents."""
         if not self.family_documents:
             return FamilyStatus.CREATED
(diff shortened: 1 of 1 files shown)


app/db/models/law_policy/family.py
```diff
@@ -78,6 +78,47 @@ def family_status(self) -> FamilyStatus:
         # If we get here then all must be deleted
         return FamilyStatus.DELETED
 
+    @family_status.expression
+    def family_status(cls):
```
diversemix, line 82: Need to chat with you about this - looks like there are now 2 ways to get "family_status" ?

katybaulch: I think it's only one way! The way most simple hybrid properties work is by under-the-hood converting the hybrid property Python code into SQL. We were getting a NotImplementedError when trying to filter by Family.family_status due to the doc_states list comprehension line, as self.family_documents is of InstrumentedAttribute type, meaning there are no values associated with the property - it's just the relationship being returned not the actual family documents. Adding the hybrid expression in conjunction with the hybrid property for our use case (where SQL alchemy and Python seeming struggle to convert our hybrid property code to the SQL required) means that we can retrieve family_status and use it in DB filter queries, as it will determine the list of family documents rather than only the relationship via the class level rather than the instance level.

katybaulch: Happy to chat more about this?

PR #185, opened 2023-11-27
Added Industry Report as submission type

katybaulch:
Added Industry Report submission type for OEP docs (hack ingest).


THOR300 (approved):
Did the ingest fail then to warrant this Pr?

PR #186, opened 2023-11-28
Feature/pdct 614 handle vespa errors gracefully (Part Two)

olaughter:
Initial feedback for api users when running a query with an invalid query string. The query error has more details so we could send that back in the future too if we decide it is a good idea to share it.

This builds on work in the data access library: https://github.com/climatepolicyradar/data-access/pull/102 The version of the data access library needs updating here after that is deployed.

To Do
- [x] Bump DAL version (Ready, see: https://github.com/climatepolicyradar/navigator-backend/pull/189)

# Description

Please include:
- a summary of the changes

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

There are no vespa tests to build on, we need to come back and add some. It is a low risk change though as vespa is not live yet

## Reviewer Checklist

- [X] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/search.py
+++ b/app/api/api_v1/routers/search.py
@@ -11,7 +11,8 @@ from io import BytesIO
 from typing import Mapping, Sequence
 
 from cpr_data_access.search_adaptors import VespaSearchAdapter
-from fastapi import APIRouter, Depends, Request
+from cpr_data_access.exceptions import QueryError 
+from fastapi import APIRouter, Depends, HTTPException, status, Request
 from fastapi.responses import StreamingResponse
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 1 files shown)


app/api/api_v1/routers/search.py
```diff
@@ -68,9 +69,14 @@ def _search_request(
...
             # TODO: we may wish to cache responses to improve pagination performance
-            data_access_search_response = _VESPA_CONNECTION.search(
-                parameters=data_access_search_params
-            )
+            try:
+                data_access_search_response = _VESPA_CONNECTION.search(
+                    parameters=data_access_search_params
+                )
+            except QueryError:
+                raise HTTPException(
```
diversemix, line 77: Do we want to log this here as well?

olaughter: Maybe, there is already some logging done as [part of the data access library](https://github.com/climatepolicyradar/data-access/blob/main/src/cpr_data_access/search_adaptors.py#L102-L107).  Do you have thoughts on wether we need anything further?

app/api/api_v1/routers/search.py
```diff
@@ -68,9 +69,14 @@ def _search_request(
...
             data_access_search_params = create_vespa_search_params(db, search_body)
             # TODO: we may wish to cache responses to improve pagination performance
-            data_access_search_response = _VESPA_CONNECTION.search(
-                parameters=data_access_search_params
-            )
+            try:
+                data_access_search_response = _VESPA_CONNECTION.search(
+                    parameters=data_access_search_params
+                )
+            except QueryError:
```
THOR300, line 76: Do we not want except as e and then pass the error into the HttpException response?

olaughter: Valid question. We spoke about this and thought it might be a good future improvement

PR #187, opened 2023-11-29
Feature/add last modified field for all entities

katybaulch:
The triggers on the child entities that would update the last modified time on the related family was preventing the deletion of events. This update fixes this.


olaughter (approved):
Lets gooooo

PR #188, opened 2023-11-30
Make empty download_url consistent

olaughter:
CCLW and UNFCCC have a different approach when handling an empty value for the download_url. This makes sure that the value is set to None in UNFCCC, wether it is reading an empty string from the csv or if the field is missing

# Description

Please include:
- a summary of the changes

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

I have expanded the current tests to add data with a missing download url and ensure the output is null. As a bonus I've made it so the test output gives more details on differences

## Reviewer Checklist

- [X] No FIXMEs remain


katybaulch (commented):
Looks good, happy for you to merge when checks complete :+1:

olaughter:
| Just thinking why we would've seen the db state change so much from the cclw ingest if this is the only place this occurs?

Yeah good question, I'll do some digging. I wonder if it relates to us running a unfccc immediately before this one? Although I would have expected it to get flagged at that time...

olaughter:
Never mind, I thought I'd be able to just leave it unchanged if nothing is found but the problem wasn't just having the download_url set to "" at the start of this snippet, it's also that the value can be "" in the csv so it needs to handle both the missing field (cclw) and the missing value (unfccc)

diversemix (commented):
I need more context as to why this change is necessary? As of today these endpoints will no longer be used / removed.

olaughter:
> I need more context as to why this change is necessary? As of today these endpoints will no longer be used / removed.

This came from an issue we had in the latest ingest where it meant extra changes where getting flagged. Deleting the endpoints entirely seems like a much better solve for the issue though!

diversemix (approved):

PR #191, opened 2023-12-04
Pages Bug fix attempt no.2.

THOR300:
# Description

Addition of a method to the SearchResponse class to increment that pages on the document passage matches. 

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Unit tests that assert the method works correctly. 

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/search.py
+++ b/app/api/api_v1/routers/search.py
@@ -76,7 +76,7 @@ def _search_request(
                 data_access_search_response,
                 limit=search_body.limit,
                 offset=search_body.offset,
-            )
+            ).increment_pages()
         else:
             return _OPENSEARCH_CONNECTION.query_families(
                 search_request_body=search_body,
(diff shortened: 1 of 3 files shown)


app/api/api_v1/routers/search.py
```diff
@@ -76,7 +76,7 @@ def _search_request(
                 data_access_search_response,
                 limit=search_body.limit,
                 offset=search_body.offset,
-            )
+            ).increment_pages()
```
diversemix, line 79: I would argue that calling `process_vespa_search_reposonse` should also handle the page numbering. We are in a router here and this should not contain any search response logic. Having said that there are functions that exist at the bottom of the file with vespa logic in :cry:  -  probably a tech debt ticket

PR #192, opened 2023-12-04
Adding filter for families that are published for vespa response.

THOR300:
# Description

This is a bug fix to return from the backend api only published documents when using the vespa query configuration. 

This follows a similar flow as that for the opensearch where we filter using the published status on the family. 

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/core/search.py
+++ b/app/core/search.py
@@ -1231,8 +1231,17 @@ def process_vespa_search_response(
     vespa_search_response: DataAccessSearchResponse,
     limit: int,
     offset: int,
+    document_extra_info: Mapping[str, Mapping[str, str]],
 ) -> SearchResponse:
     """Process a Vespa search response into a F/E search response"""
+
+    vespa_search_response.families = [
+        family
+        for family in vespa_search_response.families
+        if family.id in document_extra_info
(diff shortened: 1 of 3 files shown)


THOR300:
Just create a dummy mapping

app/core/search.py
```diff
@@ -1231,8 +1231,17 @@ def process_vespa_search_response(
     vespa_search_response: DataAccessSearchResponse,
     limit: int,
     offset: int,
+    document_extra_info: Mapping[str, Mapping[str, str]],
 ) -> SearchResponse:
     """Process a Vespa search response into a F/E search response"""
+
```
diversemix, line 1237: Should this logic be on line 1118, or should we remove that comment and debug line?

THOR300: Do you mean in this function? 

`_process_vespa_search_response_families`

The reason I didn't put it in there is just so we could get the hits count. This obviously isn't a blocker however so I'm happy to refactor!

diversemix (approved):
lgtm

PR #197, opened 2023-12-11
Adding in the bugfix for the filter fields. 

THOR300:
# Description

This is a bug relating to QueryErrors being thrown when a country or other filter is used against the api. 

This fix incorporates a new version of the dal that has been updated to expect a different set of filter fields. 

These fields effectively are what is actually used in vespa. 

This means we have to update the filter fields prior to validation of the SearchParameter object. 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

- [x] Fix failing tests in dal and backend 
- [ ] Deploy to staging and test against the vespa instance (currently no vespa search tests in the backend)  

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/search.py
+++ b/app/api/api_v1/routers/search.py
@@ -53,7 +53,7 @@ search_router = APIRouter()
 def _search_request(
     db: Session, search_body: SearchRequestBody, use_vespa: bool = False
 ) -> SearchResponse:
-    if search_body.keyword_filters is not None and use_vespa == False:
+    if search_body.keyword_filters is not None and use_vespa is False:
         search_body.keyword_filters = process_search_keyword_filters(
             db,
             search_body.keyword_filters,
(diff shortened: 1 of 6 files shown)


app/api/api_v1/routers/search.py
```diff
@@ -53,7 +53,7 @@
 def _search_request(
     db: Session, search_body: SearchRequestBody, use_vespa: bool = False
 ) -> SearchResponse:
-    if search_body.keyword_filters is not None:
+    if search_body.keyword_filters is not None and not use_vespa:
```
olaughter, line 56: The double negative might read easier with: 
```suggestion
    if search_body.keyword_filters is not None and use_vespa == False:
```
What do you think?

olaughter (approved):
This looks good, nice work finding the issue. I think at some point we'll need to figure out some improvements for the data access interface to make this kind of issue less likely to catch us out, but that is definitely wider scope

PR #200, opened 2023-12-13
Allow setting `max_hits` and `limit` for vespa

olaughter:
These values where previously hardcoded, so despite this being supported by the data access library interface with vespa, and the backend api accepting parameters for these, they would nevertheless always get set to the same values.

This passes those api parameters through to the data access library, and uses the previous set values as actual upper limits —— which we still need because of limiting excessive compute and avoiding timeouts.

See notion page for more context: https://www.notion.so/climatepolicyradar/Brief-10-matches-per-doc-100-documents-per-search-8f440121f9f44907b6ab99060f757412

# Description

Please include:
- a summary of the changes

## Type of change

- [X] Bug fix
- [X] New feature
- [ ] Breaking change

## How Has This Been Tested?

Updated the relevant test to include coverage for these values

## Reviewer Checklist

- [X] No FIXMEs remain


katybaulch (approved):
Looks good on my end, but would appreciate Mark's eyes as well!

PR #204, opened 2023-12-18
Add postgres and vespa healthchecks

olaughter:
# Description

Having the backend lose connection to postgres or vespa would be bad. This adds health checks for both into the main app health check.

For postgres this takes inspiration from the admin service' implementation, which can be found here: https://github.com/climatepolicyradar/navigator-admin-backend/blob/main/app/service/health.py#L11

For vespa it makes use of the built in health endpoint for each instance, details here: https://docs.vespa.ai/en/reference/healthchecks.html

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


katybaulch (approved):
Lgtm 👍🏻

docker-compose.dev.yml
```diff
@@ -1,10 +1,28 @@
...
+      test: curl -f http://localhost:19071/status.html || exit 1
+      interval: 5s
+      timeout: 3s
+      retries: 30
+      start_period: 20s
+
   backend:
     environment:
       PYTHONDONTWRITEBYTECODE: 0
+      VESPA_URL: http://vespatest:19071
```
katybaulch, line 20: This environment variable needs updating in the .env.example please. Could probably also use an env variable in that file for VESPA_PORT or something as well whilst we're there.

docker-compose.dev.yml
```diff
@@ -1,10 +1,28 @@
...
 services:
+
+  vespatest:
+    image: vespaengine/vespa:8.273.20
+    container_name: vespatest
+    ports:
+      - 8080:8080
+      - 19071:19071
+    healthcheck:
+      test: curl -f http://localhost:19071/status.html || exit 1
```
katybaulch, line 11: Is there a reason we're referring to this via localhost rather than the intended container name (like we do in the `docker-compose.yml` file)?

katybaulch: Typically, we've been discarding the stdout of the healthcheck e.g., `curl -s -f http://localhost:19071/status.html >/dev/null || exit 1`. Not sure whether that might be doing here?

katybaulch (changes requested):
Please update the VESPA_URL value in the .env.example file to match what is necessary for the test environment. Looking good other than that - just a few minor comments but nothing else stopping this going in 👍🏻

docker-compose.dev.yml
olaughter: Good spot, Updating this now

docker-compose.dev.yml
olaughter: Thanks Katy, updating the .env.example now

olaughter: Thanks for the comment on the port as well, after mulling it over, I'd be keen to leave the port as is as for now. It's only needed in dev so I _think_ we'd end up having to add extra complexity in prod/staging to avoid having it in those environments. Also, looking at the other services in .env.example, it seems appending the port directly to the url is consistent with opensearch, etc too, so i'd also rather go for consistency for now and potentially revisit all of them together separate of this change if we decide we want to

katybaulch: Happy with this!

katybaulch (approved):
All happy assuming tests pass 🦖

PR #205, opened 2023-12-19
Switch searches to use vespa by default

olaughter:
# Description

Makes vespa the primary search database and transfers opensearch tests over to run against vespa. This does not delete the opensearch tests or code as we're planning on keeping them around a bit longer while we build confidence in vespa.

Points to note:
- At this stage the switchover is done by changing the use_vespa parameter to default to true, this is the path we have tested and should mean opensearch can still be queried until we come back and delete it
- Test setup and coverage could be better, I refactored the opensearch versions somewhat but had to draw a line in the sand
- Having a copy of the vespa schema in this repo is less than ideal. I initially attempted to fetch the production config as part of test setup but it's in a private repo, which leads to more issues than its worth. Hoping to revisit this problem to find a better solution at some point, otherwise we'd risk schema drift.
- Some tests could not be copied over, either because they no longer are needed (yay!), or because of outstanding issues (boo!).

## Next steps:
- [x] Ensure ingest into Opensearch and Vespa is complete
- [ ] Setup and test vespa prod credentials
- [ ] Merge, semvar and deploy this branch

## Type of change

- [x] Bug fix
- [X] New feature
- [x] Breaking change

## How Has This Been Tested?

I rewrote as many opensearch tests as I could. Also months of iterating and manual testing

## Reviewer Checklist

- [X] No FIXMEs remain


tests/routes/test_vespasearch.py
```diff
@@ -0,0 +1,743 @@
...
+def _parse_id(schema, convert_to: Optional[str] = None) -> str:
+    schema_id = schema["id"].split("::")[-1]
+    if convert_to is None:
+        return schema_id
+    else:
+        id_parts = schema_id.split(".")
+    return f"{id_parts[0]}.{convert_to}.{id_parts[2]}.{id_parts[3]}"
+
+
+def _get_family_fixture(doc):
```
THOR300, line 63: Return types and type hints for parameters in some of these?

tests/routes/test_vespasearch.py
```diff
@@ -0,0 +1,743 @@
...
+    with open(VESPA_DOCUMENT_PATH, "r") as vd:
+        documents = json.load(vd)
+
+    for doc in documents:
+        family = _get_family_fixture(doc)
+        yield doc, family
+
+
+def _populate_search_db_families(db: Session) -> None:
+    run_data_migrations(db)
```
THOR300, line 80: Just to confirm this is running migrations against rds not the search database - if so naming?

tests/routes/test_vespasearch.py
```diff
@@ -0,0 +1,743 @@
...
+    db.commit()
+    db.add(family_document_slug)
+    db.commit()
+    db.refresh(family_document)
+
+
+def _make_search_request(client, params):
+    response = client.post(SEARCH_ENDPOINT, json=params)
+    assert response.status_code == 200
+    return response.json()
```
THOR300, line 259: Above this point in the module there's a lot of setting up rds and util functions. Thoughts on splitting out and making it very clear what's happening?

tests/conftest.py
```diff
@@ -62,6 +65,19 @@ def test_opensearch():
     yield connection
 
 
+@pytest.fixture(scope="session")
```
THOR300, line 68: Nice!

tests/routes/test_vespasearch.py
```diff
@@ -0,0 +1,743 @@
...
+        }
+        body = _make_search_request(client, params)
+
+        time_taken = body["total_time_ms"]
+        times.append(time_taken)
+
+    average = sum(times) / len(times)
+    assert average < REASONABLE_LATENCY_MS
+
+    with open(f"/data/benchmark_{label}_vespa.txt", "w") as out_file:
```
THOR300, line 347: How does this work then when thinking about local vs ci? 
If I run tests 4 times during development I'll have my own 4 entries do I commit these or are we adding to a `.gitignore`, then when I push to CI an uncommitted change will be appended as well?

tests/routes/test_vespasearch.py
```diff
@@ -0,0 +1,743 @@
...
+
+    # Search
+    params = {"query_string": "and", "year_range": year_range}
+    body = _make_search_request(client, params=params)
+    assert len(body["families"]) > 0
+
+    # Browse
+    params = {"query_string": "", "year_range": year_range}
+    body = _make_search_request(client, params=params)
+    assert len(body["families"]) > 0
```
THOR300, line 535: Do we need to interrogate the family events to confirm the year?

tests/routes/test_vespasearch.py
```diff
@@ -0,0 +1,743 @@
...
+        "query_string": query,
+        "keyword_filters": {
+            "countries": ["south-korea"],
+            "sources": ["CCLW"],
+            "categories": ["Legislative"],
+        },
+        "year_range": (1900, 2020),
+    }
+
+    _ = _make_search_request(client, params)
```
THOR300, line 573: No assertion?

THOR300 (commented):
Hi! Great to get this ticked off!

Left some none blocking comments. Just some thoughts and a pointer as to whether we need docstrings and return types etc. 

I'm just holding off approving as this is such an important and large PR.

The question I'm thinking is "All these tests look very good and concise but how do we know we've tested everything we need to test prior to a switch" - does copying OS tests leave us exposed to anything vespa specific etc.?

tests/routes/test_vespasearch.py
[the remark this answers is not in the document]
olaughter: Good shout, done

tests/routes/test_vespasearch.py
olaughter: fixed

tests/routes/test_vespasearch.py
olaughter: Yep, nice idea

tests/routes/test_vespasearch.py
olaughter: All good, everything under the data subfolder is ignored by the gitignore

tests/routes/test_vespasearch.py
olaughter: This one currently just tests that the request is successful (within `_make_search_request`), definetly opportunity to improve this test, but I'll leave it for today

tests/routes/test_vespasearch.py
olaughter: I'm going to ticket this along with your other comment, it's a good shout but I'm thinking not one for today

makefile-docker.defs
```diff
@@ -72,12 +72,52 @@ build:
...
+		exit 1; \
+	fi
+
+vespa_healthy:
+	@if [ ! $$(curl -f -s 'http://localhost:19071/status.html') ]; then \
+		echo 'ERROR: Bad response from local vespa cluster, is it running?'; \
+		exit 1; \
+	fi
+
+VESPA_TEST_APP_DIR='tests/search_fixtures/vespa_test_schema'
```
diversemix, line 88: I'm not sure if you're use of variables in a makefile is correct and is actually doing what is intended. Just a flag to make sure we are using the right type of the two flavours of vars: https://www.gnu.org/software/make/manual/html_node/Flavors.html

olaughter: At one point this was being used by more than one target, but it actually doesnt need to be a variable anymore, so I can sidestep this 😁

olaughter: Done

diversemix (approved):
Awesome work - lots of effort - much appreciated you did the heavy lifting on this :bow:

tests/routes/test_search.py
```diff
@@ -273,6 +273,7 @@ def test_slug_is_from_family_document(test_opensearch, client, test_db, monkeypa
             "limit": 2,
             "offset": 0,
         },
+        params={"use_vespa": False},
```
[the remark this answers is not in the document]
olaughter: Ah thats nice. I'm hoping we can delete this file in a few weeks, but good shout for next time!

THOR300 (approved):
LGTM

PR #211, opened 2024-01-16
Add tests to covering disparity between databases

olaughter:
# Description

For an upcoming pipeline infrastructure change, we needed to know how the backend might behave if there is data in just one of our two databases. This adds two exploratory tests, one covering when there is a document in postgres but not vespa, and the other checking the reverse — in vespa but not postgres. 

Both scenarios still return the correct documents, and a 200 response code. However, it turns out there is some impact on the backend for each:

### When Postgres has a document but Vespa does not
- The document will still appear in browse (which seems fine)

### When Vespa has a document but Postgres does not
- The hits will still be returned based on vespa content as vespa assumes postgres will find the data. There is potential for this to impact pagination as the frontend uses this number to determine how many pages to display.
- The backend logs errors that look like this: `ERROR    app.core.search:search.py:1116 Could not locate family with import id 'CCLW.family.8633.0'`

The option 'When Postgres has a document but Vespa does not', seems acceptable, and conveniently is how the pipeline would currently be setup.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/tests/routes/setup_search_tests.py
+++ b/tests/routes/setup_search_tests.py
@@ -68,18 +68,25 @@ def _fixture_docs() -> Iterable[tuple[VespaFixture, VespaFixture]]:
         yield doc, family
 
 
-def _populate_db_families(db: Session) -> None:
+def _populate_db_families(db: Session, max_docs: int = 4) -> None:
+    """
+    Sets up the database using fixtures
+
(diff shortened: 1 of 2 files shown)


diversemix (approved):
Looking good - like the approach

PR #214, opened 2024-01-24
Pdct 788/redirect to cdn instead of streaming response

katybaulch:
# Description

- Updated download route to redirect to CDN link instead of returning a streaming response
- Fixed bug where blank file was being uploaded to S3 due to not resetting buffer position
- Return 404 if we can't find the file at the absolute CDN path
- Converted SQL query to use Python string concatenation rather than multi-line string that seemed to be causing empty results
- Added time range to query to only dump documents that were last modified (or created) during all previous ingets cycles, so we exclude the current ingest cycle
- Linear ticket [here](https://linear.app/climate-policy-radar/issue/PDCT-788/redirect-to-cdn-instead-of-streaming-response)

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/search.py
+++ b/app/api/api_v1/routers/search.py
@@ -200,66 +200,29 @@ def download_all_search_documents(db=Depends(get_db)) -> RedirectResponse:
 
     s3_document = S3Document(DOC_CACHE_BUCKET, AWS_REGION, data_dump_s3_key)
     if valid_credentials is True and (not s3_client.document_exists(s3_document)):
-        _LOGGER.info("Redirecting to create data dump route...")
-        redirect_url = f"{PUBLIC_APP_URL}/api/v1/searches/create-all-search-dump"
-        return RedirectResponse(
-            redirect_url,
-            # status_code=status.HTTP_307_TEMPORARY_REDIRECT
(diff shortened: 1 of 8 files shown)


app/api/api_v1/routers/search.py
```diff
@@ -189,44 +188,51 @@ def download_all_search_documents(db=Depends(get_db)) -> StreamingResponse:
...
-    filename = f"whole_database_dump-{timestamp}.csv"
-    return StreamingResponse(
-        content=BytesIO(s3_file.read()),
-        headers={
-            "Content-Type": "text/csv",
-            "Content-Disposition": f"attachment; filename={filename}",
-        },
-    )
+    s3_document = S3Document(DOC_CACHE_BUCKET, AWS_REGION, data_dump_s3_key)
+    if s3_client.document_exists(s3_document):
```
diversemix, line 228: Really, really pedantic point - but prefer the "if" statements to be the exceptions to the expected flow. i.e. `if not s3_client.document_exists(s3_document): raise ... `

diversemix (approved):
lgtm :+1:

PR #216, opened 2024-01-25
Feature/pdct 774 remove csvs from backends pipeline trigger

olaughter:
# Description
Thanks to the admin service we no longer need to pass csv's into the pipeline trigger endpoint. This means a whole swathe of code is no longer used. Unpicking this means we can simplify the trigger and no longer run some unnecessary and cumbersome tests (A minute and a half! Which is half the unit tests!).

Unfortunately, pulling on the code-deletion thread revealed a place the old ingest code was in fact still used in two places:
- A handful of minor components where used for the app, so have been preserved
- The whole process was actually being used for test setup

The test setup one presents a problem. But also points to a wider issue beyond the scope of this humble change, that we have multiple different test setup functionalities across our test suite. For now, I have simply moved the ingest code into tests (Which will make it easier to clean up and means we can still turn off the tests for it). I've also added a task to our linear board to [consolidate all our test setup strategies](https://linear.app/climate-policy-radar/issue/PDCT-799/consolidate-test-setup-in-the-backend) (and move away from inline csv data fixtures 🙀) so that this can be flollowed up on in the future

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


app/api/api_v1/routers/cclw_ingest.py
diversemix, line 1: Worth renaming this file? `pipeline_trigger.py` ?

app/main.py
```diff
@@ -132,7 +131,6 @@ async def root():
 # Create an admin router that is a combination of:
 admin_router = APIRouter()
 admin_router.include_router(cclw_ingest_router)
```
diversemix, line 133: yeah this would be good to be renamed

diversemix (approved):
lgtm - the suggested rename can always get done as part of the next stage of tidy up :crossed_fingers: it happens :smile:

olaughter:
> lgtm - the suggested rename can always get done as part of the next stage of tidy up 🤞 it happens 😄

Good shout, changing now in case we never get to round two of cleanup 😅

PR #220, opened 2024-02-06
Remove opensearch code

olaughter:
Having moved to using Vespa for search we can get rid of the Opensearch code as well as corresponding tests and documentation. If there is any more that we miss from this we can follow up with further cleanup but I think this is all of it

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Removing Opensearch

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


.env.local
diversemix, line 1: Ooo - not suggesting we change this as part of this PR. But shouln't this file be something like `.env.local.example` ?

olaughter: Probably! Although I'm not sure why we need so many either... maybe we can delete one

app/core/sensitive_query_terms.tsv
diversemix, line 1: Ooo - what are we dong with sensitive terms in Vespa?

olaughter: We do this in the data access library now: https://github.com/climatepolicyradar/data-access/blob/main/src/cpr_data_access/search_adaptors.py#L85

diversemix: Now here https://github.com/climatepolicyradar/data-access/blob/main/src/cpr_data_access/search_adaptors.py#L85

diversemix (approved):
Looks good to me! Let's get this on staging :raised_hands:

PR #225, opened 2024-02-26
Automatically run the pipeline on a weekly schedule

olaughter:
Triggers the db state file creation endpoint every monday at 9:15. This means we can run during work hours for the time being as we grow in confidence in the pipeline run, later it can be moved to run overnight instead.

The implementation is intended to be simple and is likely to be moved to a seperate service from the backend in the future.

@diversemix pointed me towards [fastapi-scheduler](https://github.com/amisadmin/fastapi-scheduler), I ended up using [apscheduler](https://github.com/agronholm/apscheduler) instead as it is what the former depends on for scheduling, so we can remove some complexity. Happy to be challenged on this if there might be a reason I missed to stick with the former.

# Description

Please include:
- a summary of the changes

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

I tested the cron functionality on a cron job that ran every few seconds and just outputted a message.

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/initial_data.py
+++ b/app/initial_data.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import logging
 
 from http.client import OK
 import os
@@ -10,6 +11,9 @@ from db_client.initial_data import populate_initial_data
 
 from app.db.session import SessionLocal
 from app.core.security import get_password_hash
+
+
+_LOGGER = logging.getLogger(__name__)
 
 
 def wait_for_app():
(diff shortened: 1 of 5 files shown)


app/initial_data.py
```diff
@@ -25,8 +29,8 @@ def wait_for_app():
                 return
             else:
                 print(f"Health code: {response.status_code}, retry: {i}/{attempts}")
-        except requests.ConnectionError:
-            pass
+        except requests.ConnectionError as e:
+            _LOGGER.warning(f"Retrying health check following error: {e}")
```
diversemix, line 33: nice

diversemix (approved):
Amazing - can't wait till next Monday!

PR #232, opened 2024-03-13
Make filters exclusive to each other

olaughter:
This surfaces the changes from the data access library: https://github.com/climatepolicyradar/data-access/pull/141

As part of that it also fixes an issue with country & region searches, where country codes from both would be combined into one, so we'd always get the whole region, rather refining within it.

Note that there was two functions handling this logic, one for search and another for browse (which had no tests), I've partially consolidated them but had to add a workaround until we come back and do this properly.

# Description

Please include:
- a summary of the changes

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/search.py
+++ b/app/api/api_v1/routers/search.py
@@ -11,6 +11,7 @@ from io import BytesIO
 from typing import Mapping, Sequence
 
 from cpr_data_access.exceptions import QueryError
+from cpr_data_access.models.search import filter_fields
 from cpr_data_access.search_adaptors import VespaSearchAdapter
 from fastapi import APIRouter, Depends, HTTPException, Request, status
 from fastapi.responses import StreamingResponse
@@ -30,13 +31,13 @@ from app.core.config import (
     VESPA_URL,
 )
 from app.core.download import create_data_download_zip_archive
-from app.core.lookups import get_countries_for_region, get_country_by_slug
 from app.core.search import (
     ENCODER,
     FilterField,
     create_vespa_search_params,
     process_result_into_csv,
     process_vespa_search_response,
+    _convert_filters,
 )
 from app.db.session import get_db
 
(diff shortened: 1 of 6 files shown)


app/api/api_v1/routers/search.py
```diff
@@ -245,28 +246,18 @@ def process_search_keyword_filters(
     db: Session,
     request_filters: Mapping[FilterField, Sequence[str]],
 ) -> Mapping[FilterField, Sequence[str]]:
+    filters = _convert_filters(db, request_filters)
+    if not filters:
+        return None
```
THOR300, line 251: Does the return type of the function need changing if the response can be None?

tests/core/test_search.py
```diff
@@ -256,51 +258,170 @@ def test_create_vespa_search_params(
...
+            {FilterField.REGION: ["north-america"], FilterField.SOURCE: ["CCLW"]},
+            {"family_source": ["CCLW"], "family_geography": ["CAN", "USA"]},
+        ),
+        ({FilterField.CATEGORY: ["Executive"]}, {"family_category": ["Executive"]}),
+        ({FilterField.LANGUAGE: ["english"]}, {"document_languages": ["english"]}),
+        ({FilterField.SOURCE: ["CCLW"]}, {"family_source": ["CCLW"]}),
     ],
 )
-def test__convert_filters(test_db, filters):
+def test__convert_filters(test_db, filters, expected):
```
THOR300, line 420: Is that a typo __

tests/core/test_search.py
```diff
@@ -256,51 +258,170 @@ def test_create_vespa_search_params(
...
+            },
+            None,
+        ),
+        ({FilterField.REGION: ["north-america"]}, {"family_geography": ["CAN", "USA"]}),
+        (
+            {
+                FilterField.REGION: ["north-america"],
+                FilterField.COUNTRY: ["not-a-country"],
+            },
+            {"family_geography": ["CAN", "USA"]},
```
THOR300, line 376: Is this correct based off of the and condition? I.e. shouldn't we be returning None as the country doesn't exist in north america?

THOR300 (approved):
LGTM

app/api/api_v1/routers/search.py
olaughter: Yep, thanks for flagging, I'll get it on the next sweep

tests/core/test_search.py
olaughter: Current behaviour is to ignore countries that don't exist, theres not a need for anything better than this right now because the frontend only allows specific values. Not sure what we want longer term, maybe an error message saying if a country filter doesnt exist, but I think it needs more thought

tests/core/test_search.py
olaughter: Nope, the function is `_convert_filters`

PR #233, opened 2024-03-13
Reduce warnings from dependencies when running tests

olaughter:
# Description

Fast api & fastapi-pagination where also emitting some, as was pytest on some invalid config. Updating them has seemed to fix the issue. I tried to remove torch as it was emitting a warning and we aren't using it directly, it'll take a bit more work to remove though as it is used by other dependencies. However, I have removed torchvision so thats a start at least.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN if [ "$(wc -l <requirements.txt)" = "0" ] ; \
     fi
 
 # Install torch-cpu with pip
-RUN pip3 install --no-cache "torch==2.0.0+cpu" "torchvision==0.15.1+cpu" -f https://download.pytorch.org/whl/torch_stable.html
+RUN pip3 install --no-cache "torch==2.0.0+cpu" -f https://download.pytorch.org/whl/torch_stable.html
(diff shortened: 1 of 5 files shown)


requirements.txt
katybaulch, line 1: Do we still need a requirements.txt file for this repo?

Dockerfile
katybaulch, line 1: is there a reason the dockerfile is still using it rather than doing a `RUN poetry config virtualenvs.create false && poetry install --no-directory ` ?

katybaulch: & Can we add the requirements.txt to the gitignore too please if we don't need it

katybaulch (approved):
lgtm

Dockerfile
katybaulch: PDCT-969

olaughter: Because its silly, and yes good shout

PR #235, opened 2024-03-19
Pdct 885/add api endpoint for returning all of the geographies and their stats

katybaulch:
# Description

Add skeleton for `/geographies` endpoint in the nav-backend to be used by the world map.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN apt update && \
 
 # Install pip and poetry
 RUN pip install --no-cache --upgrade pip
-RUN pip install --no-cache "poetry==1.3.2"
+RUN pip install --no-cache "poetry==1.6.1"
 
 # Create layer for dependencies
 # TODO: refine this as part of CI & test updates
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN apt update && \
 
 # Install pip and poetry
 RUN pip install --no-cache --upgrade pip
-RUN pip install --no-cache "poetry==1.6.1"
+RUN pip install --no-cache "poetry==1.7.1"
 
 # Create layer for dependencies
 # TODO: refine this as part of CI & test updates
--- a/app/api/api_v1/routers/geographies.py
+++ b/app/api/api_v1/routers/geographies.py
@@ -21,10 +21,9 @@ async def geographies(db=Depends(get_db)):
             "iso_code": "AFG",
             "slug": "afghanistan",
             "family_counts": {
-                "executive": 100,
-                "legislative": 50,
-                "unfccc": 25,
-                "other": 0,
+                "Executive": 100,
+                "Legislative": 50,
+                "UNFCCC": 25,
             },
         },
         {
@@ -32,10 +31,9 @@ async def geographies(db=Depends(get_db)):
             "iso_code": "BGD",
             "slug": "bangladesh",
             "family_counts": {
-                "executive": 1000,
-                "legislative": 500,
-                "unfccc": 250,
-                "other": 0,
+                "Executive": 1000,
+                "Legislative": 500,
+                "UNFCCC": 250,
             },
         },
     ]
--- a/app/api/api_v1/routers/geographies.py
+++ b/app/api/api_v1/routers/geographies.py
@@ -1,0 +1,41 @@
+import logging
+
+from fastapi import APIRouter, Depends
+
+from app.api.api_v1.schemas.geography import Geography
+from app.db.session import get_db
+
+_LOGGER = logging.getLogger(__file__)
+
+geographies_router = APIRouter()
+
+
+@geographies_router.get("/geographies", response_model=list[Geography])
+async def geographies(db=Depends(get_db)):
+    """Get a summary of all geographies for world map."""
+    _LOGGER.info("Getting detailed information on all geographies")
+
+    return [
+        {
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN apt update && \
 
 # Install pip and poetry
 RUN pip install --no-cache --upgrade pip
-RUN pip install --no-cache "poetry==1.3.2"
+RUN pip install --no-cache "poetry==1.6.1"
```
olaughter, line 12: I'm all for this (1.3.2 is like over a year old!), but this does makes this inconsistent with most of our other repos so might need attention as a separate bit of work, is it needed as part of this change?

olaughter: Also also! Any reason (if/when we do start changing these) not to go straight to the latest, just checked and there is a 1.8.2 available

app/main.py
```diff
@@ -1,27 +1,27 @@
...
+from db_client import run_migrations
 from fastapi import APIRouter, Depends, FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi_health import health
 from fastapi_pagination import add_pagination
-from apscheduler.schedulers.background import BackgroundScheduler
 from starlette.requests import Request
-from contextlib import asynccontextmanager
-
-from db_client import run_migrations
```
olaughter, line 15: Yay Isort!! Are you interested in adding this to pre commits? Probably separately to this change though as I expect its going to have a lot to say for the codebase as a whole

app/api/api_v1/routers/geographies.py
```diff
@@ -0,0 +1,39 @@
...
+
+geographies_router = APIRouter()
+
+
+@geographies_router.get("/geographies", response_model=list[Geography])
+async def geographies(db=Depends(get_db)):
+    """Get a summary of all geographies for world map."""
+    _LOGGER.info("Getting detailed information on all geographies")
+
+    return [
```
olaughter, line 18: This looks good as a skeleton to allow the frontend to start using it, but do you have details on the next steps? Thinking that if we deploy this to production we can probably bet that no one will find the api link, but probably wouldn't want it there as a skeleton for too long all the same

app/api/api_v1/schemas/geography.py
```diff
@@ -0,0 +1,14 @@
...
+Json = dict[str, Any]
+
+
+class Geography(BaseModel):
+    """DTO for Geography object used by world map."""
+
+    display_name: str
+    iso_code: str
+    slug: str
+    family_counts: Json
```
olaughter, line 14:
```suggestion
    family_counts: Mapping[str, int]
```

olaughter (commented):
Looking good! Couple of questions below. I'd also be keen to see a basic test added for the endpoint at this stage, it would especially be useful when coming back and making the data dynamic. Something as simple as calling the endpoint and ensuring a response would be sufficient at this stage I think?

The [pipeline ingest test](https://github.com/climatepolicyradar/navigator-backend/blob/main/tests/routes/test_pipeline_trigger.py#L18) uses a `test_client` that could be used for this. I'm not sure its a great model for a test overall, but it might be a useful reference still

Dockerfile
katybaulch: Hey, I was trying to build the Docker image but it was failing due to the Poetry version being out of step with the lockfile. I went with 1.6.1 as that's what is in CI for most repos and seems to be the version fixed in Dockerfiles in the other repos I'm regularly in - but I have no preference on whether we should be on this, 1.7.1 or higher/latest!

app/main.py
katybaulch: It's rolled in as part of the Trunk config! I have a separate ticket for adding trunk to this repo and fixing autofixable errors, but as I have trunk (& ruff, black etc) setup locally that's why this has snuck in. More on this next week in lunch and learn! 
https://linear.app/climate-policy-radar/issue/PDCT-986/add-trunk-to-nav-backend

Dockerfile
katybaulch: The current precommits are checking for 1.7.1, so I'm going to bump this to 1.7.1 for now in this PR. We definitely need to standardise this across repos though - will likely add this as a custom precommit check in Trunk

olaughter: Sounds good, and looking forward to it!

olaughter: Just occured to me (in case you werent already planning to do this) that it might be worth double checking the image size after this change, its probably fine but we have some slightly hacky image size optmisations in the dockerfile so itd be worth ensuring they still work at later versions

olaughter (approved):

app/api/api_v1/routers/geographies.py
katybaulch: Fred and I have spoken in Slack about this in more detail and he is happy with the current approach (as the next PR for this ticket will be raised immediately after this has merged). He's happy for me to resolve these conversations in the meantime

Dockerfile
katybaulch: 2.13GB :)

katybaulch: Which is very similar to before, so no worries here! Fred and I have spoken in Slack about this in more detail and he is happy for me to resolve these conversations in the meantime

PR #237, opened 2024-03-20
Add continuation token interface for requests

olaughter:
Adds the ability to send and receive continuation tokens to navigate query grouping results (family sets, passage sets).

Adds the DAL functionality that was added in https://github.com/climatepolicyradar/data-access/pull/145 to the search api.

Note that this does not change the way we currently paginate searches. We still ask vespa for 100 results everytime and then slice it with the limit/offset values in the backend as a post-processing step. Further changes will be required on this depending on how we want to use continuations.

Also Note! Browse requests don't use vespa so don't yet support any kind of continuation tokens. This will also need addressing separately

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


app/api/api_v1/schemas/search.py
```diff
@@ -68,7 +68,8 @@ class SearchRequestBody(BaseModel):
     limit: int = 10  # TODO: decide on default
     offset: int = 0
 
-    continuation_token: Optional[str] = None
+    continuation_tokens: Optional[Sequence[str]] = None
+    continuation_token: Optional[str] = None  # Deprecated
```
katybaulch, line 72: Hey hey, mind ELI5ing this for me? How come we need to keep it in if it's deprecated?

katybaulch (approved):
Looks good! one question about the continuation_token schema with the deprecated comment next to it, but otherwise all good. Happy if you want to wait for a second pair of eyes 🦖

app/api/api_v1/schemas/search.py
olaughter: It's probably not necessary, I wasn't sure if it was being used so thought I'd do all the additive changes here and then do all the breaking changes to the api as a [separate PR](https://github.com/climatepolicyradar/navigator-backend/pull/238)

PR #240, opened 2024-03-21
Move cron job to startup

olaughter:
We had the problem where three of these jobs started at the same time,
likely because there where three backend instances running at the time
due to the backend being at its scaling limit. The thinking is when the
parent process forks it is also brings the cron job with it.

The hope here is that moving it to the startup will cause it to only run
once. As an extra check this also adds an id and sets the job to replace
any existing jobs with that id.

It's hard to test this to know if it will work for sure, but it is
already broken, and if it turns out this still doesnt work I'll just
switch back to triggering it manually until there is enough time to do a
better solution

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -73,6 +73,17 @@ async def lifespan(app_: FastAPI):
     """Run startup and shutdown events."""
     logger.info("Starting up...")
     run_migrations(engine)
+    scheduler = BackgroundScheduler()
+    scheduler.add_job(
+        start_scheduled_ingest,
+        "cron",
+        id="ingest_trigger",
+        day_of_week="mon",
+        hour=9,
+        minute=15,
+        replace_existing=True,
+    )
+    scheduler.start()
     yield
     logger.info("Shutting down...")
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -73,6 +73,17 @@ async def lifespan(app_: FastAPI):
     """Run startup and shutdown events."""
     logger.info("Starting up...")
     run_migrations(engine)
+    scheduler = BackgroundScheduler()
```
THOR300, line 76: I'm fairly certain this won't save us as if another instance is bought up this startup lifespan event will still be triggered!

olaughter: Does that mean we run migrations every time the scaling forks a new node? For what its worth Im not super confident this will save us either but didn't have time to switch to something better this week, If it doesnt work on Monday my plan would be to just remove it and hopefully will be able to replace it in prefect by the next one

THOR300: From my understanding - yes; we run alembic upgrade head on startup (this is pre the db client work) and thus if we are at the correct revision there would just be no updates. 

I'll approve so we can give this a go and me be proven wrong haha!

THOR300 (approved):
LGTM

app/main.py
olaughter: 🤞

PR #246, opened 2024-03-27
Feature/pdct 965 parameter handling needs deduplicating

olaughter:
Searches are inefficient because we validate requests with a backend pydantic model, and then load the content into a data access library one. This will have an impact on query time and also has impacted the delivery speed of recent changes. This simplifies the request as much as possible. There is two outstanding pieces of unique backend behaviour, filter processing and the limit/offset pagination. These can be revisited in the future and for now extending the data access library's model still allows us to only use one model rather than two.

This should have no impact on the frontend as I've made sure to support previous naming for requests.

This is entirely dependent on: https://github.com/climatepolicyradar/data-access/pull/157 once it is merged I will come back and update the dependency here so its not pointing at the specific branch

Also! We have the same problem with the response! but separating them felt like the best way of minimising how messy this change needs to be.

## Type of change

- [x] Refactor
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ python = "^3.9"
 Authlib = "^0.15.5"
 bcrypt = "^3.2.0"
 boto3 = "^1.26"
-cpr-data-access = {git = "https://github.com/climatepolicyradar/data-access.git", tag="v0.5.3", extras = ["vespa"]}
+cpr-data-access = {git = "https://github.com/climatepolicyradar/data-access.git", tag="v0.5.6", extras = ["vespa"]}
 fastapi = "^0.104.1"
 fastapi-health = "^0.4.0"
 fastapi-pagination = { extras = ["sqlalchemy"], version = "^0.12.19" }
--- a/tests/core/test_search.py
+++ b/tests/core/test_search.py
@@ -18,15 +18,10 @@ from sqlalchemy.orm import Session
 
 from app.core.config import VESPA_SEARCH_MATCHES_PER_DOC, VESPA_SEARCH_LIMIT
 from app.core.search import (
-    FilterField,
     SearchRequestBody,
-    SortField,
-    SortOrder,
     create_vespa_search_params,
     process_vespa_search_response,
     _convert_filters,
-    _convert_sort_field,
-    _convert_sort_order,
 )
 from tests.core.ingestion.helpers import populate_for_ingest
 
@@ -66,7 +61,7 @@ def db_setup(test_db):
             True,
             None,
             None,
-            SortOrder.ASCENDING,
+            "asc",
             None,
             10,
             10,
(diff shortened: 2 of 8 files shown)


pyproject.toml
```diff
@@ -9,7 +9,7 @@ python = "^3.9"
 Authlib = "^0.15.5"
 bcrypt = "^3.2.0"
 boto3 = "^1.26"
-cpr-data-access = {git = "https://github.com/climatepolicyradar/data-access.git", tag="v0.5.3", extras = ["vespa"]}
+cpr-data-access = {git = "https://github.com/climatepolicyradar/data-access.git", branch="feature/pdct-965-parameter-handling-needs-deduplicating", extras = ["vespa"]}
```
katybaulch, line 12: Eeeek - can this be a tag please??

katybaulch (changes requested):
Looking good - just need to update the data access library version & regenerate the poetry.lock file 👍🏻

pyproject.toml
olaughter: Yep will do! I hadn't merged the other PR yet which is all it was waiting for

tests/core/test_search.py
```diff
@@ -66,7 +61,7 @@ def db_setup(test_db):
             True,
             None,
             None,
-            SortOrder.ASCENDING,
+            "asc",
```
diversemix, line 64: I don't understand why we are now using magic strings and not constants?

diversemix: Ah ok penny dropped - this is test data - and we want to emulate the frontend :)

tests/routes/test_vespasearch.py
```diff
@@ -259,39 +257,33 @@ def test_specific_doc_returned(test_vespa, monkeypatch, data_client, data_db):
...
     query_spy = mocker.spy(search._VESPA_CONNECTION, "search")
 
     _make_search_request(
         data_client,
         params={
-            "query_string": params.query_string,
-            "exact_match": params.exact_match,
-            "limit": params.limit,
-            "max_hits_per_family": params.max_hits_per_family,
+            "query_string": "the",
```
diversemix, line 275: Have we lost the ability to do an "exact match" search?

pyproject.toml
olaughter: Done!

diversemix (approved):
Looks good  - just the query on the test ?

tests/routes/test_vespasearch.py
olaughter: We only lost it from this test. I realised the limits where the only things that still needed coverage from what this test was testing previously after consolidating the model, so I converted this test. The exact match parameter wasn't really adding anything to the test coverage here so I just removed it for simplicity. (It still works as a query type though and we have test coverage for exact matches elsewhere)

katybaulch (approved):
nice, thanks!!

PR #257, opened 2024-04-04
Make test types run concurrently

olaughter:
Currently if vespa search tests fail we need to rerun unit tests. Or if we fail unit tests we don't know if there are also issues with search tests. This wastes time and is annoying. Here we run both in parallel instead. This should save a minute or two in total anyway, but even better, means we can rerun each in isolation

The downside of this change is we now will spend longer building the image as we do it for both job, but we can (and should) follow up on this with some layer caching)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


.github/workflows/ci.yml
```diff
@@ -69,9 +92,27 @@ jobs:
         if:   always()
         run:  docker-compose logs
 
+  integration-tests:
+    runs-on: ubuntu-latest
+    steps:
       - name: Run Integration Tests
         run: echo TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO
 
+  push-to-ecr:
```
katybaulch, line 101: Can we rename this `build` please for consistency with the new CI/CD pipeline job names?

olaughter: Will do

.github/workflows/ci.yml
```diff
@@ -13,38 +13,29 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       - name: Run tests
         run:  make test_bashscripts
 
-  build-start-test:
+  unit-tests:
```
katybaulch, line 21: Needs to use `make test_unit` instead of `make test`

.github/workflows/ci.yml
```diff
@@ -69,9 +92,27 @@ jobs:
         if:   always()
         run:  docker-compose logs
 
+  integration-tests:
+    runs-on: ubuntu-latest
+    steps:
       - name: Run Integration Tests
         run: echo TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO-TODO
```
katybaulch, line 99: Should run `make test`

katybaulch (changes requested):
Just a couple of changes to the make commands we're using to run the tests please 🙏🏻 This is a fab addition though - so much easier to debug!

.github/workflows/ci.yml
```diff
@@ -13,38 +13,29 @@ jobs:
...
-        env:
-          VESPA_CLI_VERSION: "8.250.43"
-        run: |
-          mkdir scripts/vespa-cli
-          curl -fsSL https://github.com/vespa-engine/vespa/releases/download/v${VESPA_CLI_VERSION}/vespa-cli_${VESPA_CLI_VERSION}_linux_amd64.tar.gz | \
-            tar -zxf - -C scripts/vespa-cli --strip-component=1
-          echo "scripts/vespa-cli/bin" >> $GITHUB_PATH
-
       - name: Build
         run: |
```
katybaulch, line 33: Needs to have a `docker cp .env.example .env` before we build

.github/workflows/ci.yml
olaughter: Called non-search following discussion on slack

.github/workflows/ci.yml
olaughter: Agreed this needed a bigger refactor, which is also in the works (discussion on slack)

katybaulch (approved):
Nice, thanks!! 🦖

PR #276, opened 2024-04-30
PDCT 1085/corpus include text image

diversemix:
# Description

- Return corpus image and text as part of /config
- use main branch while auto-tag is being fixed
- bump minor version

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/core/organisation.py
```diff
@@ -39,12 +40,15 @@ def get_organisation_taxonomy(db: Session, org_id: int) -> Taxonomy:
 
 
 def _to_corpus_data(row, event_types) -> CorpusData:
+    image_url = f"https://{config.CDN_DOMAIN}/{row.image_url}"
```
olaughter, line 43: Two little thoughts on this. Firstly, is there a chance that the image url could start with a `/`? in which case the url would end up invalid? If so, `urljoin`, [could be an option](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin).  Related to that, is wether its worth having some simple url validation here as well? The same library might help with that too: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

olaughter (approved):
lgtm! One thought below on a small risk

app/core/organisation.py
diversemix: Ah the test is a fixed (valid) url - so I;m happy for now - going to merge but more changes to come so do shout if you want to continue the conversation

PR #282, opened 2024-06-03
Add a healthier vespa health check

olaughter:
# Description

Previous version ensured the instance had been deployed, but not that we could query it. Running an actual query checks for bad credentials, as pyvespa will raise an error on response codes between 400 and 599. The query in used here would work regardless of schema, and as it asks for 0 hits runs extremely fast

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "navigator_backend"
-version = "1.12.4"
+version = "1.12.5"
 description = ""
 authors = ["CPR-dev-team <tech@climatepolicyradar.org>"]
 packages = [{ include = "app" }, { include = "tests" }]
--- a/app/core/health.py
+++ b/app/core/health.py
@@ -1,17 +1,13 @@
 import logging
 
-import requests
 from fastapi import Depends
-from requests.compat import urljoin
 from sqlalchemy.orm import Session
 
-from app.core.config import DEVELOPMENT_MODE, VESPA_URL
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "navigator_backend"
-version = "1.12.4"
+version = "1.12.5"
```
THOR300, line 3: Is this required or do we not have auto tagging?

diversemix (approved):
lgtm

pyproject.toml
olaughter: It seems currently, yes, because the pre commit validator will fail otherwise:
https://github.com/climatepolicyradar/check-pyproject-package-version/blob/main/action.yml#L50-L51
Called by:
https://github.com/climatepolicyradar/reusable-workflows/blob/a43ba21a96bfdb1c6ee05283a9a8e7b50edd63b1/.github/workflows/python-precommit-validator.yml#L31-L32
Which is called by:
https://github.com/climatepolicyradar/navigator-backend/blob/main/.github/workflows/ci-cd.yml#L36

PR #283, opened 2024-06-04
Fix download all endpoint

olaughter:
# Description

Fixes the download all query, and adds a test.

This query broke following schema changes. This updates the sql to work for the new schema. It also adds an initial test that checks endpoint successfully returns a redirect url. This makes sure the endpoint works. Data validation would be a sensible next step but is beyond the scope here.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


app/core/download.py
```diff
@@ -189,10 +189,12 @@ def create_query(ingest_cycle_start: str) -> str:
         "    ON d.family_import_id = f.import_id "
         "inner join geography g "
         "    on g.id = f.geography_id "
-        "join family_organisation fo "
-        "    on fo.family_import_id = f.import_id "
+        "join family_corpus fc "
```
THOR300, line 192: I opened this file to read the surrounding SQL and it's rather complex and long! 

Do you think this should be managed in a different way?

THOR300 (approved):
LGTM - just the query re the sql!

app/core/download.py
olaughter: Maybe, but not as part of this change

olaughter: I'm not sure there is an ideal solution for this complex of a query, like we could move it to use sqlalchemy's ORM, but it's likely to be messier and harder to read ¯\_(ツ)_/¯

diversemix: Half way house may be to pop it in a file and read it in - so we can test it manual if needs be /cc @katybaulch

katybaulch: Yeah, Peter's approach is likely best.

PR #286, opened 2024-06-26
Simplify dockerfile

olaughter:
The previous version of the dockerfile saught to solve a size problem caused by torch. Torch is an implicit dependency from the sdk and is required for embeddings. The GPU version of torch is huge. And would break app runner if we tried to include it, we only need the cpu version so we hard coded the installation for that in the dockerfile.

Aside from looking a little messy, this made it difficult to install dependencies outside of docker, and managing dependencies this way leaves extra room for versioning issues.

This moves the torch dependency to poetry so it can be more explicitly, and more consistently, managed. Note that this requires accounting for mac specific torch install behaviour.

I built the image locally and the size did not increase in size, i want to double check this is true in the container registry too. Would also appreciate reviewers confirming this doesnt interfere with their local setup.

# Description

Please include:

- a summary of the changes

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [x] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


diversemix:
I currently have test failures locally - looking at this now

diversemix (approved):
Looks good - although we have found some weird issues with "make test"

PR #301, opened 2024-07-30
Ignore db_client when doing dependency updates

katybaulch:
# Description

Ignore db_client when doing dependency updates [PDCT-1232](https://linear.app/climate-policy-radar/issue/PDCT-1232/automate-dependency-updates-for-navigator-backend)

The db client versioning needs to be closely controlled at the moment as not every new release of the db client is always ready to propagate through into both the nav backend and admin backend at the same time (one release might remove the references in the admin service as we want to check things still work before propagating through to the nav backend for example).

Because of this, we want to force Dependabot to ignore the db_client package so we can manually control this. The db client isn't published to PyPi, just GitHub.

@THOR300 is the point of reference for dependabot - so we need a review from him before we merge.

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,7 @@ updates:
       interval: monthly
     ignore:
       - dependency-name: cpr_sdk
+      - dependency-name: db_client
   - package-ecosystem: pip
     directory: /
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,13 +6,14 @@ updates:
       interval: monthly
     ignore:
       - dependency-name: cpr_sdk
+      - dependency-name: db-client
   - package-ecosystem: github-actions
     directory: /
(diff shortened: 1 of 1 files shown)


THOR300 (commented):
I think you need the dep to be named "db-client" as this is what's declared in poetry.

I think you also need to add it for ignore in the pip package ecosystem entry as well!

THOR300 (approved):
LGTM

PR #315, opened 2024-08-27
Use only the FamilyGeography abandon geography_id on family

diversemix:
# Description
This sees all the backend code no longer referencing the `geography_id` on family

Additional manual tests:
* Locally I have removed the column from the model (using a branch in db-client) - this was useful to ensure I had caught all instances of its reference.
* For the download whole database - I now have split the query out into its own file - and I tested on a local db from a production snapshot having removed the column. I also captured the output of the query before and after to check there were no differences.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/db/crud/geography.py
```diff
@@ -14,6 +14,39 @@
...
+        db.query(
+            func.min(Geography.value).label("value"),
+            func.min(Geography.slug).label("slug"),
+            FamilyGeography.family_import_id,
+        )
+        .join(FamilyGeography, FamilyGeography.geography_id == Geography.id)
+        .filter(FamilyGeography.family_import_id == Family.import_id)
+        .group_by(Geography.value, Geography.slug, FamilyGeography.family_import_id)
+    )
+    """ NOTE: This is an intermeadiate step to migrate to multi-geography support.
```
katybaulch, line 31: Typo - should be
```suggestion
    """ NOTE: This is an intermediate step to migrate to multi-geography support.
```

app/db/crud/geography.py
```diff
@@ -14,6 +14,39 @@
...
+    We grab the minimum geography value for each family to use as a fallback for a single geography.
+    This is beacause there is no rank for geography values and we need to pick one.
```
katybaulch, line 33: Typo - should be
```suggestion
    This is because there is no rank for geography values and we need to pick one.
```

app/db/crud/geography.py
```diff
@@ -14,6 +14,39 @@
...
+            func.min(Geography.value).label("value"),
+            func.min(Geography.slug).label("slug"),
+            FamilyGeography.family_import_id,
+        )
+        .join(FamilyGeography, FamilyGeography.geography_id == Geography.id)
+        .filter(FamilyGeography.family_import_id == Family.import_id)
+        .group_by(Geography.value, Geography.slug, FamilyGeography.family_import_id)
+    )
+    """ NOTE: This is an intermeadiate step to migrate to multi-geography support.
+    We grab the minimum geography value for each family to use as a fallback for a single geography.
```
katybaulch, line 32: What do we mean by minimum in this context? The first geography value a family has?

katybaulch (approved):
lgtm - just a couple of typos :)

app/db/crud/geography.py
diversemix: ah yes

PR #317, opened 2024-08-28
Use sdk model for pipeline input

olaughter:
The sdk & backend both had the same identical model. This makes changes more convoluted and runs the risk of growing divergence over time. Here we consolidate them to use the sdk version instead, meaning that new fields can be managed in one place.

Note that there was the small config difference already which I have preserved by inheriting the sdk model into the existing backend model. I wanted to minimise how much I was changing all in one go but we can take this further in the future if we would like to.

See [all sdk changes](https://github.com/climatepolicyradar/cpr-sdk/compare/v1.1.9...v1.3.11) that are introduced here. (Essentially various version bumps + the model changes). And the [code for the sdk model](https://github.com/climatepolicyradar/cpr-sdk/blob/main/src/cpr_sdk/pipeline_general_models.py#L16) replacing the backend model.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [x] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/schemas/document.py
+++ b/app/api/api_v1/schemas/document.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from typing import Any, Mapping, Optional, Sequence, Union
 
+from cpr_sdk.pipeline_general_models import BackendDocument
 from pydantic import BaseModel, ConfigDict, field_validator
 
 from app.api.api_v1.schemas import CLIMATE_LAWS_MATCH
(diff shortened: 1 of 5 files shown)


katybaulch (approved):
ughhh beautiful tidy, have all the stars ⭐ 🌟

tests/unit/app/core/test_pipeline.py
```diff
@@ -12,10 +12,9 @@ def test_generate_pipeline_ingest_input(data_db: Session):
 
     documents = generate_pipeline_ingest_input(data_db)
     assert len(documents) == 2
-    assert documents[0].name == "Fam1"
-    assert documents[0].import_id == "CCLW.executive.2.2"
-    assert documents[1].name == "Fam1"
-    assert documents[1].import_id == "CCLW.executive.1.2"
+    for d in documents:
+        assert d.name == "Fam1"
+        assert d.import_id in ["CCLW.executive.2.2", "CCLW.executive.1.2"]
```
diversemix, line 17: The logic here is wrong - as it would pass if it returned one of these documents twice

diversemix: (I corrected the logic in the PR that just got merged)

app/api/api_v1/schemas/document.py
```diff
@@ -1,6 +1,7 @@
 from datetime import datetime
 from typing import Any, Mapping, Optional, Sequence, Union
 
+from cpr_sdk.pipeline_general_models import BackendDocument
```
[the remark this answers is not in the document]
olaughter: Adding hte details mentioned on slack here: Yep, the pipeline uses the same model and yes the Sdk model is already updated ready for you

tests/unit/app/core/test_pipeline.py
olaughter: Great spot, thanks Peter!

PR #326, opened 2024-09-10
build: Match CPR SDK `torch` version

jesse-c:
# Description

See: https://github.com/climatepolicyradar/cpr-sdk/blob/a4768cb88cf99e6b4ea87325ab3c0d60f721710c/pyproject.toml#L43

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,12 +35,7 @@ db-client = { git = "https://github.com/climatepolicyradar/navigator-db-client.g
 urllib3 = "<2"
 apscheduler = "^3.10.4"
 numpy = "1.26.4"
-torch = [
-  { platform = "darwin", url = "https://download.pytorch.org/whl/cpu/torch-2.0.0-cp310-none-macosx_10_9_x86_64.whl", markers = "platform_machine=='x86_64'" },
-  { platform = "darwin", url = "https://download.pytorch.org/whl/cpu/torch-2.0.0-cp310-none-macosx_11_0_arm64.whl", markers = "platform_machine=='arm64'" },
-  { platform = "linux", version = "2.0.0", source = "pytorch" },
-  { platform = "win32", version = "2.0.0", source = "pytorch" },
-]
+torch = ">= 2.0.0, <= 2.2.2"
 python-dateutil = "^2.9.0.post0"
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
katybaulch, line 1: Ooo - I thought we needed some of those binary files because our dockerfiles for local development weren't being built correctly without them?

pyproject.toml
```diff
@@ -35,12 +35,7 @@ db-client = { git = "https://github.com/climatepolicyradar/navigator-db-client.g
 urllib3 = "<2"
 apscheduler = "^3.10.4"
 numpy = "1.26.4"
-torch = [
-  { platform = "darwin", url = "https://download.pytorch.org/whl/cpu/torch-2.0.0-cp310-none-macosx_10_9_x86_64.whl", markers = "platform_machine=='x86_64'" },
-  { platform = "darwin", url = "https://download.pytorch.org/whl/cpu/torch-2.0.0-cp310-none-macosx_11_0_arm64.whl", markers = "platform_machine=='arm64'" },
-  { platform = "linux", version = "2.0.0", source = "pytorch" },
-  { platform = "win32", version = "2.0.0", source = "pytorch" },
-]
+torch = ">= 2.0.0, <= 2.2.2"
```
olaughter, line 38: Unfortunately, this will default to using the GPU version, which bloats the image size to more than what apprunner can handle :/

pyproject.toml
olaughter: It is indeed, and in turn a fall out of trying to only install the cpu version — which saves ~8gb in the image size :/

jesse-c:
@katybaulch, @olaughter: You made good points! I'll ditch this, especially since we're getting close to removing PyTorch.

PR #327, opened 2024-09-11
Surfacing mcf required filters in the backend 

THOR300:
# Description

This pull request surfaces changes made to the `cpr_sdk` to support filtering for the `MCF` corpus data. 

- Rename a data access legacy naming convension. 
- Add family geographies to the `SearchResponseFamily` object. 
- Update the default behaviour for which fields we filter on in vespa from `geography` to `geographies`.
- Bump the `cpr_sdk` version to contain the new queries. 
- Update the test setup so we have the option to use deterministic metadata. This is to enable the metadata test. 
- Update tests. 

### Note:
We can't test for the `corpus_type_name` or `corpus_import_id` in the search response as it isn't included in the object. 

Do we want this as part of this PR or as part of future work? 
- This would mean accessing it in the hit from vespa and adding it to the search response object. 

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/api/api_v1/schemas/search.py
```diff
@@ -251,6 +251,11 @@ class SearchResponseFamily(BaseModel):
     The geographical location of the family in ISO 3166-1 alpha-3
     """
 
+    family_geographies: List[str]
+    """
+    The geographical locations of the family in ISO 3166-1 alpha-3
```
jesse-c, line 256: **suggestion:**
```suggestion
    The geographical locations of the families in ISO 3166-1 alpha-3
```

app/core/search.py
```diff
@@ -391,6 +391,7 @@ def _process_vespa_search_response_families(
                 or hit.family_category is None
                 or hit.family_source is None
                 or hit.family_geography is None
```
jesse-c, line 393: **question:** If this was replaced above, could it be removed from here?

jesse-c (commented):
Looks reasoanble to me. I don't think I have enough understanding of this service though!

app/api/api_v1/schemas/search.py
THOR300: So in this context it's we have a search response for a family and thus singular!

olaughter (commented):
Functionally this looks great to me, and I think the tests you've added look good. I am wondering if we might need one more test in the search tests that makes sure we get metadata back? Like how we are currently doing for geographies. This would give us good confidence in this all fitting together correctly. This is good to go from my view otherwise!

THOR300:
A thought - add multiple geographies to test data to ensure we're ready for that as I think they're all just a list of the single geo currently.

THOR300:
Have you spun up the backend locally outside of tests and ran a query?

THOR300:
> Functionally this looks great to me, and I think the tests you've added look good. I am wondering if we might need one more test in the search tests that makes sure we get metadata back? Like how we are currently doing for geographies. This would give us good confidence in this all fitting together correctly. This is good to go from my view otherwise!

Definitely!

app/core/search.py
THOR300: I'm not sure I follow! This is a check as to whether the `hit` that is returned from vespa has all the relevant information to construct a `SearchResponseFamily` object.

tests/search/setup_search_tests.py
```diff
@@ -75,7 +78,10 @@ def _populate_db_families(db: Session, max_docs: int = VESPA_FIXTURE_COUNT) -> N
         if doc["fields"]["family_document_ref"] not in seen_family_ids:
             _create_family(db, family)
             _create_family_event(db, family)
-            _create_family_metadata(db, family)
+            if not deterministic_metadata:
```
olaughter, line 81: Do tests fail if we just make all of them deterministic?

tests/search/setup_search_tests.py
```diff
@@ -182,6 +188,24 @@ def _create_family_metadata(db: Session, family: VespaFixture):
...
 
 
+def _create_family_metadata_deterministic(db: Session, family: VespaFixture):
+    metadata_values = defaultdict(list)
+    family_metadata = family["fields"]["metadata"]
+    if not family_metadata:
+        return
+    for metadata in family_metadata:
+        name = metadata["name"]  # type: ignore
+        value = metadata["value"]  # type: ignore
```
olaughter, line 198: Wondering what the linting issue is that is being ignored here? If we definitely need it it'd be good to add the reason

olaughter: suggestion: Not sure if this makes a linting difference, but super minor simplification if you are interested
```suggestion
        name, value = metadata.values()
```

olaughter (approved):
Looking good! Some minor points to make decisions on below. Nothing blocking though, good work! 👍 👍

tests/search/setup_search_tests.py
THOR300: image

For some reason the linter thinks metadata is a string

tests/search/setup_search_tests.py
THOR300: Yes there was like ~9 failures so given the time I just went with this approach!

olaughter: Sounds reasonable 👍

tests/search/setup_search_tests.py
olaughter: Ah this is probably a failing of `VespaFixture`

olaughter: Recommend leaving it then in that case, we can resolve it another time

PR #351, opened 2024-09-24
Use origin as custom app configuration audience on decode

katybaulch:
# Description

Fix staging - use origin from CORS as custom app configuration token audience value on decode.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/search.py
+++ b/app/api/api_v1/routers/search.py
@@ -15,6 +15,7 @@ from cpr_sdk.search_adaptors import VespaSearchAdapter
 from fastapi import APIRouter, Body, Depends, Header, HTTPException, Request, status
 from fastapi.responses import StreamingResponse
 from jwt import PyJWTError
+from pydantic_core import Url
 from sqlalchemy.orm import Session
 from starlette.responses import RedirectResponse
 
@@ -116,7 +117,6 @@ def search_documents(
             }
         ),
     ],
-    host: Annotated[str, Header()],
     app_token: Annotated[str, Header()],
     db=Depends(get_db),
 ) -> SearchResponse:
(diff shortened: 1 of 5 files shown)


app/api/api_v1/routers/search.py
```diff
@@ -140,19 +140,23 @@ def search_documents(
     the search database. The continuation token can be used to get the next set of
     results from the search database. See the request schema for more details.
     """
+    origin = request.headers.get("origin")
+    if origin is not None:
```
diversemix, line 144: Assume if origin is None then we'll throw on 159?

diversemix: We should perhaps test this as part of `test_decoding_token_with_invalid_aud_raises_expired_signature_token_error` ?

diversemix (approved):
Looks good - but let's pop in that test ?

app/api/api_v1/routers/search.py
katybaulch: If the audience is None here, decode_config_token will throw an `InvalidAudienceError` which is captured under `test_search_decoding_token_raises_PyJWTError`

diversemix: That test confirms the http response given a thrown error. Not ... if the origin is none the error is thrown

Or am I missing something?

katybaulch: That's fair, I'll patch that in a new test :)

PR #355, opened 2024-09-25
Remove verify_aud check when in DEVELOPMENT_MODE

katybaulch:
# Description

- If on staging, set DEVELOPMENT_MODE to True
- Otherwise see if DEVELOPMENT_MODE is set, default to False if not
- Add verify_aud to jwt.decode options and set its bool value on the negation of the value of DEVELOPMENT_MODE
- Update tests

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -20,6 +20,10 @@ INGEST_CYCLE_START = os.getenv("INGEST_CYCLE_START")
 DOC_CACHE_BUCKET = os.getenv("DOCUMENT_CACHE_BUCKET")
 PIPELINE_BUCKET: str = os.getenv("PIPELINE_BUCKET", "not_set")
 INGEST_TRIGGER_ROOT: Final = "input"
-DEVELOPMENT_MODE: bool = os.getenv("DEVELOPMENT_MODE", "False").lower() == "true"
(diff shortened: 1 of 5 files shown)


diversemix (commented):
I don't think we can set DEVELOPMENT_MODE to True in staging - in `app/core/aws.py` it is used as a switch to determine if a session token in used to access S3 (line 69)

diversemix (approved):
lgtm

katybaulch:
> I don't think we can set DEVELOPMENT_MODE to True in staging - in `app/core/aws.py` it is used as a switch to determine if a session token in used to access S3 (line 69)

You're absolutely right, we can't :( will go with your original solution of completely removing the audience check

PR #370, opened 2024-10-08
Feature/pla 216 surface concept filtering changes in the backend

THOR300:
# Description

This pull request surfaces concept filtering from the `cpr_sdk` into the backend to enable filtering on concepts by the main navigator frontend / backend system. 

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


tests/search/vespa/test_vespa_concept_filters.py
```diff
@@ -0,0 +1,86 @@
...
+        ("search", "the", [{"name": "name", "value": "environment"}]),
+        (
+            "search",
+            "the",
+            [
+                {"name": "parent_concept_ids_flat", "value": "Q0,"},
+                {"name": "id", "value": "0"},
+            ],
+        ),
+        (
```
THOR300, line 27: @katybaulch I'm getting some odd behaviour in the tests and wondering whether anything jumps out to you. 

For this third parameterized test input I'm receiving the following error: 

> (Pdb) response 
<Response [400 Bad Request]>
(Pdb) response.text
'{"detail":"Invalid Query: Failed to build query: Invalid query parameter"}'

If I then set a query string I receive the following error: 

> >>> response 
> <Response [400 Bad Request]>
> >>> response.text
> '{"detail":"Could not decode configuration token"}'

THOR300: tests/search/vespa/test_vespa_concept_filters.py::test_concept_filters[search-the-concept_filters0] PASSED                      [ 33%]
tests/search/vespa/test_vespa_concept_filters.py::test_concept_filters[search-the-concept_filters1] FAILED                      [ 66%]

tests/search/vespa/test_vespa_concept_filters.py
```diff
@@ -0,0 +1,86 @@
...
+        {
+            "query_string": query,
+            "concept_filters": concept_filters,
+        },
+        expected_status_code=200,
+    )
+    response: SearchResponse = SearchResponse.model_validate(response)
+
+    assert len(response.families) > 0
+    for concept_filter in concept_filters:
```
THOR300, line 64: I know this is rather nested but it is a nested data structure and more readable than anything else I was thinking of refactoring to.

app/api/api_v1/routers/search.py
```diff
@@ -58,8 +58,12 @@ def _search_request(db: Session, search_body: SearchRequestBody) -> SearchRespon
     is_browse_request = not search_body.query_string
     if is_browse_request:
         search_body.all_results = True
-        search_body.documents_only = True
         search_body.exact_match = False
+
+        if search_body.concept_filters:
```
THOR300, line 63: As text is now not the only field searched on in the passages index in vespa we have to allow searching on the passages index even when the query_string is None.

THOR300:
TODO: Integrate and test this `cpr_sdk` update before merge. 
https://github.com/climatepolicyradar/cpr-sdk/pull/128/files

app/api/api_v1/routers/search.py
olaughter: I see. So this is to handle queries that are just searching for concepts? In its current form it looks like this would be quite tricky to maintain. Perhaps we could have two separate checks for each scenario, rather than a nested one? Perhaps checking first for "concept search", then an elif for browse search. What do you think?

olaughter: Appreciate its a bit muddy already. I'd much rather have an explicit flag for browse searches, rather than inferring them from empty strings. But probably not the time to address that

tests/search/vespa/test_vespa_concept_filters.py
olaughter: Haha, makes sense. Fun times!

olaughter (commented):
Looking good so far! Small thoughts

tests/search/vespa/test_vespa_concept_filters.py
jesse-c: You could use snapshot testing 😉 I'm happy to try that out quickly after this PR!

jesse-c (approved):

tests/search/vespa/test_vespa_concept_filters.py
jesse-c: With Marks help, I've got the snapshots generated!

katybaulch (approved):
lgtm!!

PR #375, opened 2024-10-09
Updating the vespa configuration.

THOR300:
# Description
 
This pull request is a bug fix for a failing local deployment of the frontend and backend. 
- Change the vespa url that the backend uses. Port `19071` is the config endpoint whereas `8080` is the content endpoint, see the documentation [here](https://docs.vespa.ai/en/vespa-quick-start.html). 
> The port 8080 is published to make the search and feed interfaces accessible from outside the docker container, 19071 is the config server endpoint. Only one docker container named vespa can run at a time so change the name if needed. See [Docker containers](https://docs.vespa.ai/en/operations-selfhosted/docker-containers.html) for more insights.
- Creates a local certificates directory with empty cert and key. The vespa code just expects us to be passing these in or an api key. 
- Adds documentation and some initial draft code for spinning up ones local environment to test the frontend and backend stack. This includes details of dumping and loading data from vespa. 

_Note: We're not actually passing in the vespa url from the `.env` into the docker-compose is there a reason for this, do we want to change this?_  

## Proposed version

- [X] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


makefile-docker.defs
```diff
@@ -3,7 +3,14 @@
 # ----------------------------------
 # starting, stopping, migrating DB
 # ----------------------------------
-start:
+
+create_mock_vespa_certs:
```
THOR300, line 7: Create the local vespa certs only if they don't exist upon `make start`

.env.example
```diff
@@ -9,7 +9,7 @@ ENABLE_API_DOCS=True
 DOCUMENT_CACHE_TTL_MS=0
 
 # Vespa config
-VESPA_URL=http://vespatest:19071
+VESPA_URL=http://vespatest:8080
```
katybaulch, line 12: Is it a good idea to maybe keep the other URL as a separate variable and a comment for each to differentiate what the URLs are for - not sure if that URL was needed for dev work by the platform team at some point?

makefile-docker.defs
katybaulch: Nice addition!

PR #376, opened 2024-10-10
test(search): Use snapshot testing

jesse-c:
# Description

To alleviate the complexity of the assertion.

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [x] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


docker-compose.dev.yml
```diff
@@ -17,7 +17,7 @@ services:
       PYTHONDONTWRITEBYTECODE: 0
       VESPA_URL: http://vespatest:19071
     volumes:
-      - ./:/cpr-backend/:ro
+      - ./:/cpr-backend/
```
jesse-c, line 20: **explanation:** Needed for writing the updated snapshots.

tests/search/vespa/test_vespa_concept_filters.py
```diff
@@ -58,29 +59,15 @@ def test_concept_filters(
...
-                            ]
-                        )
+    # Set arbitrary values, since we don't care about these changing.
+    #
+    # The snapshot testing library, syrupy, does provide other ways to achieve
+    # this, but I was struggling with getting it working at the time.
+    actual.total_time_ms = 0
+    actual.query_time_ms = 0
+
+    assert actual == snapshot()
```
olaughter, line 71: The kind of data we are using in these tests particularly lends itself to snapshots. 1. Because they are a large blob, and the change here really shows the benefits of using them, having reduced down those nested for loops into one line is really nice! 2. Because it really just speeds up how we've been handling fixtures anyway. In that we grab one copy of prod data and just check it looks like what we expect, and currently managing them when adding new fields is a faff.

olaughter: One concern that can be raised on snapshotting is that it can make it easier to miss issues. For example, if the tests are failing we might assume its just that the snapshots need updating. Do you have any thoughts on how we avoid just running the snapshot update in that situation? This could also happen with out current setup so its not at all a dealbreaker from my perspective. But would be helpful if there is something in place to mitigate even if its just a note in the readme on dealing with these

olaughter (commented):
This looks good to me! I'd be keen to see some input from @climatepolicyradar/software  before it goes in to make sure everyones on board before it potentially gets implemented in more of the test suite

diversemix (commented):
I'm aware I'm about to leave so this may be a moot point - but there are a number of reasons snapshot testing is a bad idea. Shall we discuss? /cc @jamesgorrie @olaughter

diversemix:
FTR top two bug-bears (1) False sense of security - in reality it is all to easy to save something as the correct snapshot when it indeed broken .... (2) High Maintenance and Brittle Tests - small changes can lead to an undue overhead for test maintenance

jesse-c:
No worries at all! 🙇🏻 This really was just intended as a possible quick win, and it's already required too much time for a quick win, so I'm going to close it.

PR #388, opened 2024-10-17
Make pipeline query support multi geos

katybaulch:
# Description

Make the pipeline query support multiple geographies.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


scripts/db_state_validator_click.py
```diff
@@ -0,0 +1,201 @@
...
+
+import click
+from pydantic import BaseModel, ValidationError
+
+# Setup logger
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class DocumentMetadata(BaseModel):
```
THOR300, line 14: You should be able to just import these from the sdk!

THOR300 (approved):
Great to get this in! 

@katybaulch has confirmed locally by building from this branch and on main that the produced `db_state.json` is the same barring the changes listed below. 

The only changes that were observed were: 
- any geographies changes due to now yielding all geos not just one from the geo table. 
- some changes in the slug that was produced due to a difference in ordering that was introduced. This is as we take the last slug from the family document when creating the pipeline input. 

What impact does a change in slug have on the pipeline? 
- We identify the slug update in the identify updates step. 
- We then update the json's in the pipeline cache with the new slug such that the data is correct as slugs are indexed in vespa. 
- Slugs are not used in any of the vespa queries so no search behaviour from that aspect should be affected, just some slugs in response objects (I think they are just an overhang from the original data structure and don't really actually need to be passed through the pipeline and into vespa). 

So to confirm a change to some slugs will not affect the pipeline and any geo changes are expected as we are yielding all geos from the database.

PR #426, opened 2024-12-03
Config endpoint returns only orgs that correspond to allowed corpora

annaCPR:
# Description
- require an app_token header for the /config endpoint
- validate and decode the app_token in the /config endpoint
- only return stats for organisations linked to allowed corpora from the app_token

Please include:

- a summary of the changes

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/repository/organisation.py
+++ b/app/repository/organisation.py
@@ -88,9 +88,9 @@ def get_organisation_config(db: Session, org: Organisation) -> OrganisationConfi
 
 
 def get_organisations(db: Session, allowed_corpora: list[str]) -> list[Organisation]:
-    return (
-        db.query(Organisation)
-        .join(Corpus, Corpus.organisation_id == Organisation.id)
-        .filter(Corpus.import_id.in_(allowed_corpora))
-        .all()
+    query = db.query(Organisation).join(
+        Corpus, Corpus.organisation_id == Organisation.id
     )
+    if allowed_corpora != []:
+        query = query.filter(Corpus.import_id.in_(allowed_corpora))
+    return query.all()
--- a/app/repository/organisation.py
+++ b/app/repository/organisation.py
@@ -87,5 +87,10 @@ def get_organisation_config(db: Session, org: Organisation) -> OrganisationConfi
     )
 
 
-def get_all_organisations(db: Session) -> list[Organisation]:
-    return db.query(Organisation).all()
+def get_organisations(db: Session, allowed_corpora: list[str]) -> list[Organisation]:
+    return (
+        db.query(Organisation)
+        .join(Corpus, Corpus.organisation_id == Organisation.id)
+        .filter(Corpus.import_id.in_(allowed_corpora))
+        .all()
+    )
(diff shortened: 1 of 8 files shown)


app/repository/organisation.py
```diff
@@ -87,5 +87,10 @@ def get_organisation_config(db: Session, org: Organisation) -> OrganisationConfi
     )
 
 
-def get_all_organisations(db: Session) -> list[Organisation]:
-    return db.query(Organisation).all()
+def get_organisations(db: Session, allowed_corpora: list[str]) -> list[Organisation]:
+    return (
+        db.query(Organisation)
+        .join(Corpus, Corpus.organisation_id == Organisation.id)
+        .filter(Corpus.import_id.in_(allowed_corpora))
```
katybaulch, line 94: Almost - whilst this handles the case of locking down the API based on the allowed corpora, we want to return all the results if the allowed_corpora list is == `[]`. This can be done super easily within this function luckily, simply by splitting the query up like so

```
query = (
        db.query(Organisation)
        .join(Corpus, Corpus.organisation_id == Organisation.id)
    )
if allowed_corpora != []:
    query = query.filter(Corpus.import_id.in_(allowed_corpora))
result = query.all()        
```

katybaulch: (plus any corresponding test update)

katybaulch (approved):
Great work! Minor comment based on context that I might not have made clear which is blocking - but happy as soon as that is resolved for this to go in 🍪

app/repository/organisation.py
annaCPR: Ah, yes I thought you said something like this in the morning. Could you please explain what the rationale behind this is actually? Would it not be better to error in this case?

katybaulch: It's to keep our API working. The longterm goal I believe was that we had an API that could be used by techies/researchers to access our data (as it's all open anyway). Our introduction of the app_token works for all requests generated from within a custom app, but don't work for standalone API calls when you don't know that you need to add in the app_token as a custom header - plus users can't currently generate their own custom app tokens.

tests/non_search/routers/lookups/test_config.py
```diff
@@ -285,6 +290,65 @@ def test_config_endpoint_returns_stats_for_allowed_corpora_only(
...
+        "exp": issued_at + relativedelta(years=10),
+        "iat": int(datetime.timestamp(issued_at.replace(microsecond=0))),
+        "iss": "Climate Policy Radar",
+        "sub": "CPR",
+        "aud": "localhost",
+    }
+    app_token = jwt.encode(
+        to_encode, os.environ["TOKEN_SECRET_KEY"], algorithm=security.ALGORITHM
+    )
+    # app_token = app_token_factory(",")
```
katybaulch, line 309: Commented out code

annaCPR: well spotted

katybaulch (approved):
All good once the commented out code has been removed!

PR #430, opened 2024-12-18
Feature/pdct 1759 make backend config endpoint focused on corpora not

annaCPR:
# Description

- add a new top-level corpus_types object to the /config response

- update each corpus within the corpus_types object to include org ID and name

- update count_by_category and total to count families per corpus (rather than org) and add to each corpus object in the response

- update relevant tests to check for the new structure

Please include:

- a summary of the changes

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/models/metadata.py
```diff
@@ -25,10 +25,36 @@ class OrganisationConfig(BaseModel):
...
+    organisation_id: int
+    # No of families in corpus
+    total: int
+    count_by_category: Mapping[str, int]
+
+
+class CorpusTypeConfig(BaseModel):
+    """Contains the CorpusType info as well as data of any corpora of that type"""
+
+    corpus_type_description: str
```
katybaulch, line 48: I think we need the corpus_type_name here too

app/repository/corpus.py
katybaulch, line 1: Per DDD, we want to maintain our code files split by entities (aka family, corpus, corpus_type, perhaps even config?). Additionally, the difference between service level code and repository level code is something to bear in mind (I like [this](https://stackoverflow.com/questions/4159812/services-and-repositories-in-ddd-c) stack overflow question as a quick point of reference on this). With this in mind, are there any refactor changes we could make to this file?

app/models/metadata.py
```diff
@@ -25,10 +25,36 @@ class OrganisationConfig(BaseModel):
     count_by_category: Mapping[str, int]
 
 
+class CorpusConfig(BaseModel):
```
katybaulch, line 28: Should we be moving this under a corpus or config entity file rather than leaving this in metadata (DDD)??

app/repository/corpus.py
```diff
@@ -0,0 +1,108 @@
...
+        db.query(Family.family_category, func.count())
+        .join(FamilyCorpus, FamilyCorpus.family_import_id == Family.import_id)
+        .filter(FamilyCorpus.corpus_import_id == corpus_import_id)
+        .group_by(Family.family_category)
+        .all()
+    )
+    found_categories = {c[0].value: c[1] for c in counts}
+    count_by_category = {}
+
+    # Supply zeros when there aren't any
```
katybaulch, line 30: Could we argue that this function could in fact be two repository layer functions (one for retrieving total count, one for retrieving category count), and one service layer function (the business logic for creating the family stats per corpus object)?

katybaulch: The repo layer should only really be for CRUDding for an entity/value object, whereas the service layer is for the application logic in DDD (I believe!).

katybaulch (approved):
Missing some docstrings for your new functions (sorry 😢) but otherwise all good!

PR #436, opened 2025-01-06
Feature/pdct 1567 pull corpora decoding and validation into middleware

katybaulch:
# Description

- Pull corpora decoding and validation into middleware
- Split middleware into separate file
- Skip corpora decoding and validation for all endpoints other than those specified (NEED TO THINK ABOUT HOW TO SCALE THIS GOING FORWARD)
- Update world map endpoint to return all counts when None token or no allowed corpora given

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/service/util.py
```diff
@@ -73,3 +75,17 @@ def tree_table_to_json(
...
+) -> list[str]:
+    return token.allowed_corpora_ids if token is not None else []
+
+
+def get_subject_from_token(
+    token: Optional[AppTokenFactory],
+) -> str:
+    if token is not None and token.sub is not None:
+        return str(token.sub)
+    return ""
```
jamesgorrie, line 91: Would `None` be a clearer type for "There is no subject"?

app/repository/geography.py
```diff
@@ -61,18 +60,17 @@ def get_geo_subquery(
 
 
 def count_families_per_category_in_each_geo(
-    db: Session, allowed_corpora: list[str]
+    db: Session,
+    allowed_corpora: Optional[list[str]],
```
jamesgorrie, line 64: `Optional[list[str]]` -> does "empty list" (`[]`) suffice to say "don't filter"? From the comment, it suggests so.

If that is the case, could we simplify this by just having `list[str]`?

If not - what is the difference between `[]` and `None`?

jamesgorrie (commented):
Couple of questions.

I think there is something I am missing - might be easier to chat in person, but I'll try describe my confusion.

* the middleware that sets the `raw_token` on the request
* `get_validated_token` takes that token and parses it

I can see then where `get_validated_token` is run on the world map.

`routers/documents` seems to read directly from `request.state.token` - when is this property set on the request?

PR #461, opened 2025-02-19
fix(deps): update cpr_sdk

jamesgorrie:
# Description

- updates the cdk to apply [a partial revert](https://github.com/climatepolicyradar/cpr-sdk/releases/tag/v1.16.3) giving us access back to the concepts data on the Vespa response

## Proposed version

- [x] Patch


katybaulch (approved):
Lgtm

PR #478, opened 2025-03-24
fix: use psycopg2 over psycopg2-binary

jamesgorrie:
# Description
☝️ 

[As per the docs - this is something that is recommended](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary).

Towards https://linear.app/climate-policy-radar/issue/APP-403/intermittent-production-500-errors

## Proposed version

- [x] Patch


annaCPR (approved):
lgtm

odrakes-cpr (approved):

PR #479, opened 2025-03-25
fix: update openssl in dockerfile

jamesgorrie:
# Description
- forces an update of openssl in the dockerfile

There might be a conflict in openssl version in python vs apprunner - this is trying to rule that out as an culprit for the errors we are seeing.

## Proposed version

- [x] Patch


odrakes-cpr (approved):

jamesgorrie:
Closing as it seems to be fixed by #480

PR #484, opened 2025-04-08
fix: use port 8888 over 8000

jamesgorrie:
# Description

☝️ 

I'm not sure exactly how to write a test for this? Maybe a [Docker `HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck)?

Happy to look at the testing post fix.

[This is the post used by AppRunner](https://github.com/climatepolicyradar/navigator-infra/blob/aab42de47ff7ab72832df862f25a7a26f01a230c/backend/backend_app/main.py#L153).

## Proposed version

- [x] Patch

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "navigator_backend"
-version = "1.23.17"
+version = "1.23.18"
 description = ""
 authors = ["CPR-dev-team <tech@climatepolicyradar.org>"]
 packages = [{ include = "app" }, { include = "tests" }]


odrakes-cpr (approved):

startup.sh
```diff
@@ -20,4 +20,4 @@ else
 fi
 
 echo "Starting backend app"
-exec uvicorn app.main:app --host 0.0.0.0 --port 8000
+exec uvicorn app.main:app --host 0.0.0.0 --port 8888
```
odrakes-cpr, line 23: question (non-blocking): what difference would this make

PR #488, opened 2025-04-10
Fix bug in default page ordering

katybaulch:
# Description

For some reason page 1 is being ordered at the end of the page number patches instead of the beginning? this should fix it

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/service/search.py
```diff
@@ -293,7 +293,7 @@ def process_result_into_csv(
     return csv_result_io.read()
 
 
-def _parse_text_block_id(text_block_id: Optional[str]) -> Tuple[int, int]:
+def _parse_text_block_id(text_block_id: Optional[str]) -> Tuple[Optional[int], int]:
```
jamesgorrie, line 296: fyi: new proposal for `int | None`
https://peps.python.org/pep-0604/#proposal

jamesgorrie (approved):

app/service/search.py
katybaulch: Love this! Might do it all in one go throughout the repo if that's what we're moving to though, and then I can git blame ignore the commit hash 🤔

PR #489, opened 2025-04-10
feat: deglobalisation of Vespa

jamesgorrie:
# Description
- move `_VESPA_CONNECTION` from being a module level `global` => FastApi lifecycle value
- all the refactoring that's needed for that

[Seems like a pretty standard way of doing it](https://archive.ph/Mxj4S).

[And how FastAPI do it](https://fastapi.tiangolo.com/tutorial/sql-databases/#read-one-hero).

[This is the juice of the PR](https://github.com/climatepolicyradar/navigator-backend/pull/489/commits/17532e0031aa5ca91451c3a55400bd8ad384b3d1).

This should hopefully make this value safe to share around the app without seeing the segfaults that we are currently seeing. 

And in turn remove the 500 errors we are seeing being returned from our AppRunner instance.

## Proposed version

- [x] Patch

### DIFF
--- a/makefile-docker.defs
+++ b/makefile-docker.defs
@@ -125,6 +125,9 @@ test_non_search:
 test:
 	docker compose -f docker-compose.yml -f docker-compose.dev.yml run --rm backend pytest -vvv tests ${ARGS}
 
+t:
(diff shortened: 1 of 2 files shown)


jesse-c (approved):
LGTM

PR #491, opened 2025-04-14
fix: use Depends on FastAPI route over implementation method

jamesgorrie:
# Description
- uses the [Depends](https://fastapi.tiangolo.com/tutorial/dependencies/#create-a-dependency-or-dependable) on the FastAPI route
- removes it from the method that isn't the FastAPI route as it isn't available at that point

I'll write a test post fix - but want to de-red our deployments.

## Proposed version

- [x] Patch


katybaulch (approved):
lgtm

PR #501, opened 2025-05-01
feat: families-api initial setup

jamesgorrie:
# Description

The aim of this PR is to deploy a simple service, e2e, with no hacks that haven't got described fixes. Solutions to any hacks introduces should be
- captured and referenced in Linear
- solved as separate PRs to this

So please point out any areas that we think needs work so we can get it documented.

It is also intentionally free of abstractions as I hope to discover those as we start to work on the other pieces - following the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

It is also following the principle of keeping it simple and trying to compose simple, complete abstractions into simple implementations - so not engineering for an imagined future. There is an implied pragmatism here - where we should be designing and coding with these in mind:
- security
- performance
- observability
- documentation

This is part of a larger piece of work loosely described in these phases:
- **Make it work:** `/families` => This phase, working on a low impact piece of the pie where there are no dependencies on it, but might have some value in analysis, migration etc
- **Make it good:** `/search` => ...
- **Make it fast:** `/geographies` => ...
- **Make it cheap:** `/paragraphs` => ...

https://api.climatepolicyradar.org/families/docs

## Proposed version
- [x] Patch

## End goal
```mermaid
flowchart RL
    subgraph production
        direction LR
        Rouet53["api.policyradar.io"]-->CloudFront
        CloudFront-->CloudFrontOrigin/concepts-->AppRunner/concepts-->BakedDuckDB-->ConceptsRespone
        CloudFront-->CloudFrontOrigin/families-->AppRunner/families-->RDS-->FamiliesResponse
        CloudFront-->CloudFrontOrigin/documents-->AppRunner/documents-->RDSReadReplica-->DocumentsResponse
        CloudFront-->CloudFrontOrigin/passages-->AppRunner/passages-->DataLake-->PassagesRespons
        CloudFront-->CloudFrontOrigin/labels-->AppRunner/labels-->???-->LabelsResponse
    end
```

part of: https://linear.app/climate-policy-radar/project/isolate-services-within-navigator-backend-abeb5f150aa4/overview

After an initial PR review with @annaCPR I have thought you can review this in 3 different steps:

## DevX
This is the poetry + just file. Probably interests all @climatepolicyradar/software 

## Application
Mostly very simple and non-important to this PR - we hope to get more of this sorted when introducing a new service

## Infra
This is predominantly in the `infra` folder - might interest @katybaulch @henry-cpr


families-api/README.md
```diff
@@ -0,0 +1,71 @@
...
+## TODO
+
+### Python App
+
+**these can / shoudl be added when we have another service in to ensure we get the right abstraction**
+
+- [ ] JSON logging
+- [ ] OTel implementation
+
+## Potential middle-game
```
jamesgorrie, line 44: @olaughter @henry-cpr because we've been talking about this and @katybaulch as you're doing some work with infra and we've spoken about microservices

I've tried to capture a near-end game for having a few isolated services running. They should be easy enough to spin up - and also then we have some simple things where we can start talking about things like

- "How do we use datalake for this?" (the passages app should give us a lead)
- "What about API access control" once we have the API we can try draw it into this graph e.g. Cloudwatch edge lambda on the cloudfront origins

All based on - let's get some simple things write and then start to compose around them - or, if we really really need to, add some more simple things.

families-api/infra/__main__.py
```diff
@@ -0,0 +1,165 @@
...
+    },
+    service_name="families-api",
+    source_configuration={
+        "authentication_configuration": {
+            "access_role_arn": families_api_role.arn,
+        },
+        "image_repository": {
+            "image_configuration": {
+                "runtime_environment_secrets": {
+                    "NAVIGATOR_DATABASE_URL": families_api_apprunner_navigator_database_url.arn,
```
jamesgorrie, line 157: @katybaulch might be of interest - using the secure pattern of storing values encrypted in SSM and then referencing them in the App.

families-api/justfile
```diff
@@ -0,0 +1,33 @@
+# should be used for first run
+bootstrap:
+    # TODO: find a way to ensure we're using the latest dump
+    @echo "🔄 downloading latest dump from s3"
+    aws s3 cp s3://cpr-production-rds/dumps/2025-04-23-navigator-production.sql ./initial-data/2025-04-23-navigator-production.sql
```
jamesgorrie, line 5: @odrakes-cpr - here's where I need your dump please!

families-api/infra/__main__.py
jamesgorrie: @henry-cpr you mentioned this in a previous chat too, moving away from Pulumi config.

families-api/app/main.py
```diff
@@ -0,0 +1,225 @@
+from typing import Generic, Optional, TypeVar
+
+from fastapi import APIRouter, Depends, FastAPI
+from pydantic_settings import BaseSettings
+from sqlalchemy import create_engine
+from sqlmodel import Field, Relationship, Session, SQLModel, func, select
+
+
+class FamilyCorpusLink(SQLModel, table=True):
+    __tablename__ = "family_corpus"  # type: ignore[assignment]
```
odrakes-cpr, line 10: question (non-blocking) : Wary this may be a first iteration, but I am not sure how I feel about model classes being defined in the main.py file.

families-api/app/main.py
```diff
@@ -0,0 +1,225 @@
...
+
+navigator_engine = create_engine(settings.navigator_database_url)
+
+
+def get_navigator_session():
+    with Session(navigator_engine) as session:
+        yield session
+
+
+@router.get("/", response_model=APIResponse[PhysicalDocumentPublic])
```
odrakes-cpr, line 154: quibble: as per my comments below, this feels like this should be a dedicated routes file

families-api/app/main.py
jamesgorrie: What would the benefits of this be?

Currently they are used in 1 place - so co-location feels like the simplest implementation.

These Models are also going to be incredibly simple, as services now will have 1 responsibility.

In this PR we are trying to not over-engineer up front, but rather discover the best patterns for abstractions through implementation - i.e. [Rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

families-api/app/main.py
jamesgorrie: Similar to my above comment - what are we hoping to gain from that given that this app will have a very simple router?

families-api/app/main.py
odrakes-cpr: Hmm this would just be following fast-api convention, whilst it doesn't explicitly dictate how you should define your file structure. It does suggest making things modular and [separating concerns](https://fastapi.tiangolo.com/tutorial/bigger-applications/), with the main app file minimal and business logic held elsewhere.

odrakes-cpr: But again just a quibble as you said we aren't trying to over-engineer.

jamesgorrie: Great chat - I've read that doc a million times, and think it's great!

But - the reason I am not following this is the aim of this work is to decompose our services into not-bigger-applications as described in that doc.

families-api/justfile
```diff
@@ -0,0 +1,41 @@
...
+    docker compose down
+
+dev:
+    docker compose up
+
+# build
+# TODO: this needs to be tied into a lifecycle somewhere to make sure it is run if something is added
+# sed is used here to remove a whole bunch of gumpf you can not remove with poetry v1 
+requirements:
+    poetry export --without-hashes --without-urls --format requirements.txt | sed 's/ ;.*//'  > requirements.txt
```
jamesgorrie, line 24: note: we could add poetry to the `Dockerfile`, but this maintains a slimmer `Dockerfile`, and we can introduce it later if we want.

families-api/service-manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "service.name": "families-api",
+  "service.namespace": "data-fetching",
+  "team": "application",
+  "inputs": [{ "type": "aws", "name": "navigator-infra" }],
```
jamesgorrie, line 5: Make the input the RDS data store

families-api/app/main.py
```diff
@@ -0,0 +1,225 @@
...
+
+class APIResponse(SQLModel, Generic[APIDataType]):
+    data: list[APIDataType]
+    total: int
+    page: int
+    page_size: int
+
+
+class Settings(BaseSettings):
+    navigator_database_url: str
```
jamesgorrie, line 131: @annaCPR worth looking at as a pattern - but this is for `env_vars`

jamesgorrie: @jamesgorrie how is this working locally

families-api/app/main.py
```diff
@@ -0,0 +1,225 @@
...
+        default=None, foreign_key="corpus.import_id", primary_key=True
+    )
+    family_import_id: int | None = Field(
+        default=None, foreign_key="family.import_id", primary_key=True
+    )
+
+
+class Corpus(SQLModel, table=True):
+    __tablename__ = "corpus"  # type: ignore[assignment]
+    import_id: str | None = Field(default=None, primary_key=True)
```
henry-cpr, line 21: question: what behaviour does giving a default=None entail? And what's the logic behind nullable primary key? That feels dangerous to me

families-api/app/main.py
```diff
@@ -0,0 +1,225 @@
...
+    title: str
+    families: list["Family"] = Relationship(
+        back_populates="corpus", link_model=FamilyCorpusLink
+    )
+
+
+class FamilyGeographyLink(SQLModel, table=True):
+    __tablename__ = "family_geography"  # type: ignore[assignment]
+    geography_id: int | None = Field(
+        default=None, foreign_key="geography.id", primary_key=True
```
henry-cpr, line 31: quibble: I'm not au-fait with SQLModel, does having two primary keys make it a composite key? My personal preference is not to have composite primary keys, but to just have an auto-incrementing id field (i.e. id, geography_id, family_import_id)

families-api/app/main.py
```diff
@@ -0,0 +1,225 @@
...
+    type: str
+    slug: str
+
+
+class Geography(GeographyBase, table=True):
+    __tablename__ = "geography"  # type: ignore[assignment]
+    value: str
+    parent_id: int = Field(foreign_key="geography.id")
+    # the relationship stuff here is a little non-standard
+    # there doesn't seem to be an easier way to do this with SQLModel
```
henry-cpr, line 51: I've no issue with non-standard workarounds, but would be keen to understand if this affects / entails any lazy-loading magic or anything like that? Since we're pushing for simplicity, a good understanding of that would be useful, as many DB inefficiencies I've debugged in my time have had lazy/eager loading ORM logic root causes

families-api/infra/__main__.py
```diff
@@ -0,0 +1,313 @@
+import json
+
+import pulumi
+import pulumi_aws as aws
+
+account_id = aws.get_caller_identity().account_id
+
+
+def generate_secret_key(project: str, aws_service: str, name: str):
+    return f"/{project}/{aws_service}/{name}"
```
henry-cpr, line 10: I know its implicit in terms of which SSM we're using etc, but I wonder if it's useful to put the environment in this prefix as well to reduce cognitive overhead and make it really clear when debugging

families-api/justfile
henry-cpr: I'm happy with this as is, given that logic of env management is to be able to deal with conflicting dependencies across different projects but by definition the container is its own single python environment. So this keeps principle of simplicity

henry-cpr (approved):
This is looking great. I don't have anything blocking, some minor opinions weakly held. Nice one!!

families-api/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.11-slim
```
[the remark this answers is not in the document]
jamesgorrie: Done - [with tech debt of updating the main repo too](https://linear.app/climate-policy-radar/issue/APP-594/update-python-version-in-nav-back)

families-api/infra/__main__.py
annaCPR: Is that not already included in the name of aws_service? I thought we already follow a pattern where the service name has the environment in it

families-api/app/main.py
jamesgorrie: Yea, this is quite opinionated and subjective to myself, so very loosely held.

A large source of inspiration is [Life of a file](https://www.youtube.com/watch?v=XpDsk374LDE). Ping @jesse-c for Elm funsies.

families-api/README.md
annaCPR, line 1: Can we maybe mention why we have a separate requirements.txt and that it gets autogenerated by poetry on build? This was confusing to me at first

annaCPR (approved):
Great work!

families-api/app/main.py
jamesgorrie: I've made all IDs required (copy/pasta from elsewhere)

families-api/app/main.py
jamesgorrie: Very much doing document driven development here:
https://sqlmodel.tiangolo.com/tutorial/many-to-many/create-models-with-link/#link-table-model

Letting SQLModel do it's thing until it's thing isn't good enough. Which, I think until I get your OTel observability in, is good enough.

families-api/app/main.py
jamesgorrie: This is inherited from current RDS setup - I've updated the comment as per, related to [the comment above](https://github.com/climatepolicyradar/navigator-backend/pull/501#discussion_r2088714869).

families-api/README.md
jamesgorrie: https://github.com/climatepolicyradar/navigator-backend/pull/501/commits/82d65a1275135edf81726f12754f42f2b38cecf7

Nice shout

jamesgorrie:
@henry-cpr @annaCPR @odrakes-cpr thank you for the reviews - some good stuff in there - I have been capturing things we should definitely look at in the [Linear project](https://linear.app/climate-policy-radar/project/isolate-services-within-navigator-backend-abeb5f150aa4/issues).

PR #503, opened 2025-05-12
Explicitly set poetry & pyright paths

jamesgorrie:
# Description
- explicitly sets the poetry venv path
- uses that path in pyright
- removes the need for the `configure-pyright` bootstrap

This then allows us to use the `[tool.pyright]` section of the `pyproject.toml`, which, locally, is currently being overridden by the local `pyrightconfig.json` file.

It is also more explicit and requires less steps to bootstrap the app.

## Proposed version

- [x] Patch


poetry.toml
katybaulch, line 1: This is not necessarily the virtual environment name for everyone - which is what the above script was for until we'd moved to uv. If we are still using poetry, the virtual environment name will not be `.venv`.

jamesgorrie: As per the description
- this explicitly sets this to `.venv` in the poetry.toml file
- Uses this in pyright

The idea was to standardise around this.

The bug I was actually trying to fix was - having the locally generated `pyrightconfig.json` was overriding any values in the `[tool.pyright]` section of the `pyproject.toml`.

This was creating discrepancies between local and CI which was making things really hard to debug.

Being more explicit about this and creating a standard local dev setup felt like a nice solution.

katybaulch: Thanks! My understanding of setting `venvPath` or equivalent in the `poetry.toml` file was that it sets _where_ to look for the virtual environment, but doesn't create nor guarantee that someone already has a virtual environment called `.venv` within the project workspace. Please can you clarify?? :))

jamesgorrie: 1. adding this to the `pyproject.toml` tells poetry to create the virtualenv in `./.venv`
  ```
  [virtualenvs]
  in-project = true
  path = ".venv"
  ```
2. adding `venvPath = ".venv"` to `[tool.pyright]` tells pyright to us that virtualenv

That's it.

katybaulch: Got it, thanks so much!!

PR #504, opened 2025-05-12
Fix: rewritten and efficient geo map query

henry-cpr:
# Description

This PR throws away the old geography/map query and introduces a new one that runs in THREE PERCENT of the original query time. THREE PERCENT. 

Anyway, after the fun of the incident I thought it might be a lack of indexing on the key columns in the query. @katybaulch very kindly added those, but I'm sad to report that investigating the performance post-indexing showed query times going from ~12s to ~38s, which if I remember my maths correctly is going in the wrong direction. 

Looking at the query planner, it looks like the indices shifted the query planning to an even more inefficient set of operations. Given how many repeated bits of logic, confusing implicitly materialised views, and so on the original query had, I decided it would probably be easier and quicker to re-write it from scratch. So here we go. 

I hate CASE statements (it's a declarative language lets not shoehorn procedural constructs in), so wanted to get rid of those. I also rather dislike subqueries inside CASE statements, you never know what the query planner will make of them. We do two massive joins at the end of the query on the geographies which are crazy expensive, and it's super hard to tell if we've filtered the DB down to manageable levels or not. 

Instead, I've approached as:
- Filter (ONCE) to published documents
- Use filtered documents to generate document counts 
- Use a cunning trick to build out combinatorial lookup table for all geog/category pairs
- Join that all together to match previous output format.

All investigation using full set of corpus IDs and production database dump taken this morning. 
- Original query on unindexed data:  12s
- Original query on indexed data: 38s
- Updated query in this PR: 0.421s

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [x] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

I've tested by diff'ing outputs of old query and new query on two different DB dumps which show zero diferences.
All tests passing locally. 

THAT SAID I still have a lingering anxiety that the logic is slightly different or I've caused regression somewhere, so some form of inquiry by colleagues with working brains is much appreciated. 

## Reviewer Checklist

- [] The PR represents a single feature (small driveby fixes are also ok)
- [ ] No FIXMEs remain


henry-cpr:
> Looks good, reads well! Any thoughts on testing this for release?

I was hoping for a steer to be honest. I'm still not comfortably idiomatic with navi-backend.  Once I've got CI tests working (which work locally) I think it might need to be deploy on staging and test numbers? I've diffed outputs of old query & new query showing zero difference, but its a big change so I'm loathe to call it a day there. 

I guess it's a spectrum from:
- Deploy to staging and manually have a look
to
- set up a bunch of fixtures with very disparate data distributions and run old and new queries and see if any output diffs.

Or- any magic anyone else thinks of as my brain is very low function today!

henry-cpr:
NB: trunk fmt changed the semantics of the query, making tests fail in CI that hadn't on local. 

Please challenge if preferred response, but i have disabled linting on the world map query. It's just to tide us over to the geography service...

jamesgorrie:
@henry-cpr re:testing - @odrakes-cpr had added some timing metrics to our endpoints:
https://github.com/climatepolicyradar/navigator-admin-backend/pull/287

Maybe
- load test `main` on `staging`
- deploy this to `staging`
- load test this on `staging`

Compare numbers?

katybaulch (approved):
ugh this is beautiful!!!!!! nice work!! happy with disabling the lining and seconding @jamesgorrie's comments on test plan

PR #514, opened 2025-05-15
`families-api` deploy

jamesgorrie:
# Description

- adds `staging` deploy on merge-to-main for `families-api`
- adds "[reviewed deployment](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/reviewing-deployments)" deploy to `production` AKA `latest`
- releases a hashed deploy to github SHA

## The reviewed deploy
- is on the [`production` environment](https://github.com/climatepolicyradar/navigator-backend/settings/environments/6731733887/edit)
- GitHub protection are all-or-nothing, so any other action wanting to use the production `environment` will be protected
- A mitigation for the above is creating another environment called `production-protected` or similar
- It current requires @climatepolicyradar/tech-devs to review the deploy. While I can see this being a risk when wanting to deploy in urgency (and this is already pretty much the case) - mitigations are
  - remove the gate temporarily
  - GitHub managers can bypass the gate
  - deploy locally

[Here's an example](https://github.com/climatepolicyradar/navigator-backend/actions/runs/15038351394/job/42264241207?pr=514)

## GitHub environment setup
- We have two [environments](https://github.com/climatepolicyradar/navigator-backend/settings/environments) setup in GitHub for this repo
  - `staging`
  - `production`
- each environment has an `AWS_ACCOUNT_ID` setup (currently staging is pointing to production [until we do this](https://linear.app/climate-policy-radar/project/isolate-services-within-navigator-backend-abeb5f150aa4/issues))
- this replicates what we have in `navigator-infra`

Fixes: https://linear.app/climate-policy-radar/issue/APP-575/add-deployment-for-families-api

## Proposed version

- [x] Patch

### DIFF
--- a/.github/workflows/merge-to-main.yml
+++ b/.github/workflows/merge-to-main.yml
@@ -18,7 +18,7 @@ jobs:
       - uses: aws-actions/configure-aws-credentials@v4.2.0
         with:
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/navigator-backend-github-actions-deploy
-          role-session-name: GitHub_to_AWS_via_FederatedOIDC
+          role-session-name: GitHub_to_AWS_via_FederatedOIDC_staging
           aws-region: eu-west-1
       - uses: aws-actions/amazon-ecr-login@v1
       - working-directory: families-api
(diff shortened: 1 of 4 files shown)


.github/workflows/merge-to-main.yml
```diff
@@ -0,0 +1,55 @@
+name: Merge to Main
+on:
+  push:
+    branches:
+      - main
+      - geographies-api-deploy
```
jamesgorrie, line 6: I'll remove this before merging

katybaulch (approved):
lgtm, nice work!!

PR #519, opened 2025-06-03
fix: use this concepts API for production

jamesgorrie:
# Description

- uses this repo for `concepts-api` `production`
- uses the [`initial-data` pattern from `families-api`](https://github.com/climatepolicyradar/navigator-backend/blob/main/families-api/justfile#L6) to generate the data to be baked into the `Dockerfile`
  - I'm not sure `initial-data` is as name fit for purpose, but I can change it universally later
- adds `environment` to the `justfile` as it's needed

## Bonus
- ✨ removes the `None` case from the `has_classifier` in the concepts-api as it always evaluates to `bool`  

Related: https://github.com/climatepolicyradar/concepts/pull/19
Fixes: https://linear.app/climate-policy-radar/issue/APP-643/move-infra-for-concepts-api-=-navigator-backend

## Proposed version

- [x] Skip auto-tagging

### DIFF
--- a/concepts-api/app/main.py
+++ b/concepts-api/app/main.py
@@ -287,35 +287,3 @@ async def health_check():
 
 
 app.include_router(router)
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-


jamesgorrie:
This is already deployed to production

katybaulch (approved):
big green tick -- let's get it going!!

PR #524, opened 2025-06-04
Feature/app 661 fix hardcoded classifier specsjson in concepts api

katybaulch:
# Description

- Fix hardcoded file name in concepts API for classifier specs file.
- Removed fallback to local file

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/concepts-api/create_duckdb.py
+++ b/concepts-api/create_duckdb.py
@@ -20,7 +20,12 @@ def fetch_classifier_specs_file() -> Tuple[str, bool]:
     owner = "climatepolicyradar"
     repo = "knowledge-graph"
     branch = "main"
-    path = "flows/classifier_specs/prod.yaml"
+
+    classifier_file_name = "prod.yaml"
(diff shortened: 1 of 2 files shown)


concepts-api/create_duckdb.py
```diff
@@ -31,17 +36,9 @@ def fetch_classifier_specs_file() -> Tuple[str, bool]:
         return response.text, False
```
annaCPR, line 36: question: Just for my own understanding, what is the point of returning False here (particularly as it doesn't seem like we ever return True)?

PR #546, opened 2025-06-17
String family fields

jamesgorrie:
# Description
- adds most string fields to the `FamilyPublic`

Next up: add linked values like collections etc.

## Proposed version

- [x] Skip auto-tagging


katybaulch (approved):
lgtm

PR #549, opened 2025-06-17
refactor: poetry => uv for root

jamesgorrie:
# Description

- migrates our root project to `uv` with the intention of using it to manage all microservices via [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/)

## Proposed version

- [x] Patch

### DIFF
--- a/api/infra/__main__.py
+++ b/api/infra/__main__.py
@@ -22,7 +22,17 @@ concepts_api_stack = pulumi.StackReference(f"climatepolicyradar/concepts-api/{st
 config = pulumi.Config()
 url = config.require("url")
 
-api_route53_zone = aws.route53.Zone(url, name=url)
+# as climatepolicyradar.org is still managed in the cpr AWS account
+# we need to create a whole new record for it as opposed to just an ALIAS
+
+if stack == "production":
+    api_route53_zone = aws.route53.Zone(url, name=url)
+    api_route53_zone_id = api_route53_zone.zone_id
+else:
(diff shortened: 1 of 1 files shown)


jamesgorrie:
After a chat with @katybaulch - we're waiting on the approval on the PR and then we will remove the requirement to have "`code-quality / python`" pass.

pyproject.toml
```diff
@@ -1,65 +1,63 @@
-[tool.poetry]
-name = "navigator_backend"
-version = "1.26.0"
+[project]
+name = "navigator-backend"
```
katybaulch, line 2: **question:** what's the reason for this change?

jamesgorrie: None whatsoever - just a mistake when converting it
```suggestion
name = "navigator_backend"
```

.github/workflows/ci-cd.yml
```diff
@@ -24,15 +24,19 @@ jobs:
...
-      poetry-version: "1.8.2"
-
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Check pyproject.toml package version matches
```
katybaulch, line 34: quibble: this can be removed now as we no longer have this section in the PR template

jamesgorrie: @katybaulch removed

PR #558, opened 2025-06-23
fix: return document.language as a string

jamesgorrie:
# Description

[As per the FE type - language should be a `string`](https://github.com/climatepolicyradar/navigator-frontend/blob/main/src/types/types.ts#L197).

Screenshot 2025-06-23 at 09 39 39


katybaulch (approved):
lgtmx

PR #560, opened 2025-06-23
fix: use event dates for family

jamesgorrie:
# Description

Uses the event dates for published / updated.
With fallbacks of the family data.

✨ 🧹 removes some fields from the response that aren't used by the frontend.

### DIFF
--- a/families-api/app/main.py
+++ b/families-api/app/main.py
@@ -224,8 +224,8 @@ class FamilyPublic(FamilyBase):
 
     @computed_field
     @property
-    def last_updated_date(self) -> datetime:
-        # get the most recent date that is before now
+    def last_updated_date(self) -> datetime | None:
+        # get the most recent date that is not in the future
         now = datetime.now(tz=timezone.utc)
         latest_event_date = max(
             (
@@ -235,7 +235,7 @@ class FamilyPublic(FamilyBase):
             ),
             default=None,
         )
-        return latest_event_date or self.last_modified
+        return latest_event_date
 
     @computed_field
     @property


families-api/app/main.py
```diff
@@ -206,12 +206,30 @@ def geographies(self) -> list[str]:
...
 
     @computed_field
     @property
     def last_updated_date(self) -> datetime:
-        return self.last_modified
+        updated_event = next(
+            (
+                event
+                for event in self.unparsed_events
+                if event.event_type_name == "Updated"
```
katybaulch, line 227: issue: i'm not sure this logic is right. we aren't looking for events where `event.event_type_name == "Updated"` for this, but rather any event that has the most recent associated date.

families-api/app/main.py
```diff
@@ -206,12 +206,30 @@ def geographies(self) -> list[str]:
     @computed_field
     @property
     def published_date(self) -> datetime:
-        return self.created
+        published_event = next(
+            (
+                event
+                for event in self.unparsed_events
+                if event.event_type_name == "Passed/Approved"
```
katybaulch, line 213: issue: similiarly, not all corpora have `"Passed/Approved"` as the event name that is used to calculate the published date. in the taxonomy (event taxonomy?), there is a field called `datetime_event_name` which contains the `event_type_name` that is used to calculate the published date. we should be comparing to the value of datetime_event_name (which should always be atm an array of length 1). if an `event_type_name` of `datetime_event_name[0]` does not exist anywhere in the list of events, we should default to using the event with the least recent date.

katybaulch: [Link to existing logic](https://github.com/climatepolicyradar/navigator-db-client/blob/fac0b04e70ecf1845fc729cabfaec6954319aeea/db_client/models/dfce/family.py#L209)

families-api/app/main.py
```diff
@@ -206,30 +206,29 @@ def geographies(self) -> list[str]:
...
+        # datetime_event_name stores the value of the event.event_type_name that should be used for published_date
+        # otherwise we use the earliest date
+        published_event_date = next(
             (
-                event
+                event.date
                 for event in self.unparsed_events
-                if event.event_type_name == "Passed/Approved"
+                if event.valid_metadata is not None
+                and event.event_type_name == event.valid_metadata["datetime_event_name"]
```
katybaulch, line 216: praise: nicely reflected logic & clear comments!

katybaulch (changes requested):
nice progress!  unfortunately the logic for the last_updated_date still isn't quite right. please see my original and recent comment and shout me if you want to discuss :))

katybaulch (approved):
amazing, well done!!

PR #573, opened 2025-06-30
Fix isort with microservices

katybaulch:
# Description

Since moving the backend code into its own microservice, isort has not been sorting our own modules properly - instead grouping them in with 3rd party packages. This fixes that.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


.git-blame-ignore-revs
```diff
@@ -27,3 +27,6 @@
 # Fix SQL query formatting
 047766a85f086fc0986a6f2b49fee9d73fa219e8
 ab3476708920c5760f058ec40d14d008f94f5bad
+
+# Fix isort
+b1f02734416fb7327c692e95180d25bfb315f282
```
jamesgorrie, line 32: question: I was wondering about this with my PR - but assuming we need to use the squashed PR hash once we have merged as we "Squash and merge" our PRs?

katybaulch: It'll come up in a squiggly line on main if it doesn't work - so let's give it a try and see!

PR #575, opened 2025-06-30
feat: root install

jamesgorrie:
# Description
- adds a root install to the repo
- uses uv workspaces


jamesgorrie:
In favour of https://github.com/climatepolicyradar/navigator-backend/pull/577

PR #577, opened 2025-06-30
feat: add shared lib to families-api

jamesgorrie:
# Description
- treats api as a lib to be build
- adds the api shared lib to `families-api`
- makes the `dev` docker compose use the root as the docker context

This is almost the last step to allowing us to start abstracting out things like the oTel libraries.

[Pretty much followed the `uv workspace` docs](https://docs.astral.sh/uv/concepts/projects/workspaces/).

Next is to move `concepts/geographies-api`.
I did this as the first step to keep the PR clean.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,18 +13,19 @@
 # be identified in the format @org/team-name. Teams must have
 # explicit write access to the repository.
 
-*                                                @climatepolicyradar/software
-app/api/api_v1/routers/pipeline_trigger.py       @climatepolicyradar/deng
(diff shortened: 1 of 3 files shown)


.trunk/trunk.yaml
```diff
@@ -45,7 +45,7 @@ lint:
       commands:
         - name: lint
           output: sarif
-          run: pyright --project .trunk/pyrightconfig.backend-api.json --outputjson ${target}
+          run: pyright --project .trunk/pyrightconfig.backend-api.json --outputjson
```
jamesgorrie, line 48: This was removed as it was telling `pyright` to scan the whole directory, effectively overriding the `excludes`/`includes` in the config

PR #582, opened 2025-07-01
feat: add slack for a github deploy

jamesgorrie:
# Description
- adds a new slack message on deploy to add more context of what's being deployed
- moves the service matrix to an env variable to avoid multiple deploys being triggered per service

This is a replica of [this](https://github.com/climatepolicyradar/navigator-frontend/pull/605) as I want to try out if it give us what we want i.e. more context, with no extra noise.


katybaulch (approved):
super interesting - can't wait to see it in action!! 🚀

PR #593, opened 2025-07-02
fix: update pyright config + comments

jamesgorrie:
# Description

- syncs the two configs
- describes that ./.trunk/configs/pyrightconfig.json is canonical and `[tools.pyright]` is replicated
- fixes drift

### DIFF
--- a/.trunk/configs/pyrightconfig.json
+++ b/.trunk/configs/pyrightconfig.json
@@ -1,7 +1,10 @@
 {
+  "comment-1": "this is replicated to ./pyrightconfig.json",
+  "comment-2": "if you make any changes here, please keep them updated there",
   "exclude": ["backend-api/*"],
   "include": ["api/src", "concepts-api", "families-api"],
   "venvPath": ".",
   "venv": ".venv",
-  "extraPaths": ["api/src", "concepts-api", "families-api"]
+  "extraPaths": ["api/src", "concepts-api", "families-api"],
+  "pythonVersion": "3.12"
 }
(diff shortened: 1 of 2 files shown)


.trunk/configs/pyrightconfig.json
katybaulch, line 1: question: why do we need both??

pyproject.toml
```diff
@@ -8,13 +8,14 @@ dependencies = []
 [tool.uv.workspace]
 members = ["api", "concepts-api", "families-api", "geographies-api"]
 
+# this is replicated from ./.trunk/configs/pyrightconfig.json
+# this allows bypassing Trunk for quicker feedback in the IDE
+# any changes to pyrightconfig should be made there and replicated here
```
katybaulch, line 13: issue: i think this is unnecessary. trunk will look in its configs folder, and if the config file can't be found there it'll search your project root, and then recursively until it finds a matching config file. we shouldn't need to duplicate -- see https://docs.trunk.io/code-quality/linters/supported/pyright#settings and  https://docs.trunk.io/code-quality/linters/configure-linters#moving-linters

katybaulch: You don't need to have the extension installed though - do you use the CLI standalone?

jamesgorrie: > I disable the plugin, but still want **_realtime_** feedback

"realtime" being the focus here - I expect feedback as I type, not post running a command. Happy to explore other options if we're too worried about drift?

PR #596, opened 2025-07-03
fix: remove concepts data fallback

jamesgorrie:
# Description

- removes fallback for concepts data about classifiers
- changes `List` => `list`

Fixes: https://linear.app/climate-policy-radar/issue/APP-618/remove-fallback-for-concepts-with-classifiers-to-avoid-confusion


katybaulch (approved):
lgtm:)

PR #597, opened 2025-07-03
fix: using ENV to create_duckdb for classifiers_spec file

jamesgorrie:
# Description

Uses `ENV` to calculate which [classifiers_spec](https://github.com/climatepolicyradar/knowledge-graph/tree/main/flows/classifier_specs) file to use.

[This is based on env var we use for oTel](https://github.com/climatepolicyradar/navigator-backend/blob/dfd32c2da4157cf8e61341a22d781d688f7bd14f/concepts-api/app/main.py#L24)

Fixes https://linear.app/climate-policy-radar/issue/APP-617/use-correct-env-file-from-concepts-store-to-get-concepts-with

### DIFF
--- a/concepts-api/create_duckdb.py
+++ b/concepts-api/create_duckdb.py
@@ -21,7 +21,7 @@ def fetch_classifier_specs_file() -> str:
     branch = "main"
 
     classifier_file_name = "prod.yaml"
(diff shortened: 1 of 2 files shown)


katybaulch (approved):
nice!!

PR #598, opened 2025-07-03
fix: move completely to python 3.12

jamesgorrie:
# Description

- upgrade fully to python 3.12

We might want to investigate using [`.python-version`](https://docs.astral.sh/uv/concepts/python-versions/#python-version-files) - but don't want to create more work for myself.

Fixes https://linear.app/climate-policy-radar/issue/APP-594/update-python-version-in-nav-back


katybaulch (approved):
lgtm

PR #614, opened 2025-07-16
feat: add subdivisions to geography filter query

odrakes-cpr:
# Description

We are updating our vespa query; adding subdivisions to the geography query. 

A subdivision in our code base is effectively a geography, so have updated the countries array that takes all the iso codes and passes them to vespa [see line 387].

Have added a function to validate the iso codes based on the objects in our geo database. Made the decision to use the db to validate the iso codes rather than the geo service endpoint as was concerned network latency might affect search performance.

Wonder if it is worth changing that array name from `countries` to `geographies`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/backend-api/app/service/search.py
+++ b/backend-api/app/service/search.py
@@ -377,6 +377,8 @@ def _convert_filters(
                 # TODO: remove this once frontend is updated to use ISO codes in favour of get_countries_by_iso_codes
                 get_geographies_as_iso_codes_with_fallback(db, values)
             )
+        elif field == FilterField.SUBDIVSION:
+            countries.append(get_geographies_as_iso_codes_with_fallback(db, values))
         else:
             new_values = values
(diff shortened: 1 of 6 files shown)


backend-api/app/service/search.py
katybaulch, line 1: nice! have we got any tests for this path?

odrakes-cpr: @katybaulch  this ticket is not finished and ready for review, I was just pushing to see if i broke any of the integration tests

annaCPR (approved):
lgtm 🙏  thank you

PR #634, opened 2025-07-30
fix: add import_id to FamilyEventPublic

jamesgorrie:
# Description

- adds `import_id` to FamilyEventPublic to be used by the frontend.

Everything in my head wants to use `id` - but to avoid transformation fatigue, I have stuck with this as we use it elsewhere.

CC: @jesskelsall

### DIFF
--- a/families-api/app/main.py
+++ b/families-api/app/main.py
@@ -141,7 +141,7 @@ class Collection(SQLModel, table=True):
 
 
 class FamilyEventBase(SQLModel):
-    pass
+    import_id: str
 
 
 class FamilyEvent(FamilyEventBase, table=True):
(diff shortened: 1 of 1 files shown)


katybaulch (approved):
id is a reserved keyword in python, so this is a good choice for multiple reasons! nice one

jamesgorrie:
> id is a reserved keyword in python, so this is a good choice for multiple reasons! nice one

@katybaulch Not massively idiomatic to the general API world though, `id` I would say is pretty standard.

[It doesn't seem to be an issue when used elsewhere](https://github.com/climatepolicyradar/navigator-backend/blob/e4d28eba8e3227e86f89fb650fdb144bd01e072a/families-api/app/main.py#L91)?

PR #660, opened 2025-08-13
fix: remove line length from markdown lint

jamesgorrie:
# Description

- adds the architecture diagram for `families-api`
- removes markdown lint MD013. I find myself spending an inordinate amount of time trying to fix for this which isn't time well spent. [If there was an autofix](https://github.com/DavidAnson/markdownlint/issues/294), I'd happily leave it in.

The intention here is that if we want to make large changes going forward - we can express them in the diagram first and RFC before embarking on the work.

It's also useful for new starters to get their head around it.

This might be wrong as it's my reading of it from the code, so an 🦅 👁️ would be appreciated from @annaCPR and @odrakes-cpr 

[For ease of use - here's the rendered README](https://github.com/climatepolicyradar/navigator-backend/blob/7b3058f709ac7c0fc0a045b3fd19bd44afcc1fa3/geographies-api/README.md).

### DIFF
--- a/.trunk/configs/.markdownlint.yaml
+++ b/.trunk/configs/.markdownlint.yaml
@@ -1,4 +1,25 @@
 # Autoformatter friendly markdownlint config (all formatting rules disabled)
 default: true
 
-MD013: false
+MD013:
+  # Number of characters
+  line_length: 80
+
+  # Number of characters for headings
+  heading_line_length: 80
+
+  # Number of characters for code blocks
+  code_block_line_length: 80
+
+  # Include
+  headings: true
+
+  # Don't include
+  tables: false
+  code_blocks: false
+
+  # Strict length checking
+  strict: false
+
+  # Stern length checking
+  stern: false
--- a/.trunk/configs/.prettierrc
+++ b/.trunk/configs/.prettierrc
@@ -1,0 +1,11 @@
+{
+  "overrides": [
+    {
+      "files": "*.md",
+      "options": {
+        "printWidth": 80,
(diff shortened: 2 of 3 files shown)


jamesgorrie:
re: blocking on the line length - I literally found it easier to undo than find the right length automatically.

Please can we add auto fixing as another PR - which would then autofix this readme - but I don't want it to block this.

katybaulch:
> hey -- there is a way of autofixing line length violations, it just looks like we haven't configured it. would you like me to take a look at configuring this or would you like to do it with me?

It's one very small change, which can be applied in a matter of minutes, so I'd like to vote we just do it all in one now rather than making more work for ourselves later please :)

Just need to create the following file `.trunk/configs/.prettierrc` and paste the following:

```json
{
  "overrides": [
    {
      "files": "*.md",
      "options": {
        "printWidth": 80,
        "proseWrap": "always"
      }
    }
  ]
}
```

And then our `trunk fmt` and `trunk check` both should catch this.

katybaulch:
> it feels odd to have markdown lint set the rules, and then have to replicate those in prettier to implement them

I think the 🔑 is that we should be enforcing style for consistency everywhere using the formatting tools, but we set the limits in the linter so that if for whatever reason something slips through the cracks (e.g., a commit was pushed that bypassed commit hooks that the developer forgot to go back and run the commit hooks on later), we don't end up pushing code which has regressed in terms of our own standards to `main` -- so we keep a sort of baseline.

katybaulch (approved):
lgtm, thanks for adding!

PR #678, opened 2025-08-28
feat: corpus.import_id filter on docs_by_geo

jamesgorrie:
# Description

Adds a filter similar to #677 to the `docs_by_geo`.

This should allow us to see if we have families for a geography as filtered by a custom_apps default corpuses.


katybaulch (approved):
Makes sense to me!

PR #680, opened 2025-08-28
feat: add statistics to Geography

jamesgorrie:
# Description
- adds statistics to geographies
- adds [European union](https://app.climatepolicyradar.org/geographies/european-union) to the custom countries
- makes sure we add our custom countries to the list

## example response
```json
{
    "data": {
        "id": "EUR",
        "name": "European Union",
        "statistics": {
            "name": "European Union",
            "legislative_process": "The legislative process at the European Union (EU) level involves the European Commission (independent from national governments), the European Parliament (elected by EU citizens), and the Council of the European Union, which represents Member States. Most often, the Commission proposes new legislation, but it is the Council and Parliament together that pass the laws.</div><div>The main forms of EU legislation are directives, regulations and decisions. Directives and regulations can be adopted by the Council in conjunction with the European Parliament or by the Commission alone. A regulation is a general measure that is binding in all its parts, directly appli­cable in the Member States and addressed to everyone. A directive, on the other hand, is addressed to the member states. It is binding as to the result to be achieved, but leaves member states to choose the form and method they adopt to achieve it. The Commission is required to verify that member states transpose correctly and in due time the directives that have been adopted and can sanction them if they fail to do so. Decisions are EU laws relating to specific cases. They can be adopted by the Council (sometimes jointly with the European Parliament) or by the Commission.</div><div>The Commission can also publish Action Plans, White Papers, Green Papers, Commission regulations and Communications. An Action Plan serves to detail actions needed to reach the goals set in individual directives. A White Paper sets out the Commission’s policy programme in a specific area. Before a White Paper is written, a Green Paper is published, which is a consultative document including suggestions and options for new policy. Each single proposal for legislation announced in a White Paper or deriving from a policy initiative announced in it is subject to one or more rounds of open consultation and an impact assessment. White Papers, Green Papers and Communications can serve to identify future legislative proposals. Commission regulations primarily serve as administrative acts on the functioning of the EU Institutions. The Treaty of Lisbon (2009) created a new category of legislation, Delegated Acts, by which, under strict conditions, the legislator can delegate to the Commission the power to adopt acts amending non-essential elements of a legislative act, in particular to specify certain technical details.",
            "federal": true,
            "federal_details": "27 nation states",
            "political_groups": "G20",
            "global_emissions_percent": "6.33",
            "climate_risk_index": "Not available",
            "worldbank_income_group": "High income",
            "visibility_status": "published"
        },
        "type": "country",
        "alpha_2": "EU",
        "has_subconcept": [],
        "subconcept_of": [],
        "slug": "european-union"
    }
}
```

### DIFF
--- a/geographies-api/tests/test_main.py
+++ b/geographies-api/tests/test_main.py
@@ -1,3 +1,4 @@
+import pytest
 from fastapi.testclient import TestClient
 
 from app.data.geography_statistics_by_countries import geography_statistics_by_countries
@@ -7,22 +8,27 @@ from app.model import APIItemResponse, Geography
 client = TestClient(app)
 
 
-def test_read_geography_has_statistics():
-    response = client.get("/geographies/afghanistan")
+@pytest.mark.parametrize(
+    ("slug", "expected_statistics"),
+    [
+        # We do not have statistics for regions yet
+        ("north-america", None),
+        # We do have statistics for some countries
+        ("afghanistan", geography_statistics_by_countries["AFG"]),
+        # But not all countries
+        ("holy-see-vatican-city-state", None),
+        # We do not have statistics for subdivisions yet
+        ("us-ca", None),
+    ],
+)
+def test_read_geography_statistics(slug: str, expected_statistics: dict | None):
+    response = client.get(f"/geographies/{slug}")
 
     assert response.status_code == 200
     response = APIItemResponse[Geography].model_validate(response.json())
     statistics = response.data.statistics
 
-    assert statistics is not None
-    assert statistics.model_dump() == geography_statistics_by_countries["AFG"]
-
-
-def test_read_geography_does_not_has_statistics():
-    response = client.get("/geographies/us-ca")
-
-    assert response.status_code == 200
-    response = APIItemResponse[Geography].model_validate(response.json())
-    statistics = response.data.statistics
-
-    assert statistics is None
+    if statistics is None:
+        assert expected_statistics is None
+    else:
+        assert statistics.model_dump(exclude_none=True) == expected_statistics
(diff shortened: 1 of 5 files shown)


jamesgorrie:
Create tech debt: https://linear.app/climate-policy-radar/issue/APP-1078/add-tests-for-get-geographies-service

jamesgorrie:
@annaCPR I would like to get this in so we can start deprecating things as it's causing a tonne of confusion.

Once we have 1 place of getting the data (geographies API) this will be easy to change.

annaCPR:
Can we add some tests?

jamesgorrie:
@annaCPR I've captured the tech debt and will get it in as this work is potentially unblocking for some frontend work.

geographies-api/tests/test_main.py
```diff
@@ -0,0 +1,28 @@
...
+
+    assert response.status_code == 200
+    response = APIItemResponse[Geography].model_validate(response.json())
+    statistics = response.data.statistics
+
+    assert statistics is not None
+    assert statistics.model_dump() == geography_statistics_by_countries["AFG"]
+
+
+def test_read_geography_does_not_has_statistics():
```
annaCPR, line 21: quibble: This tests only one case where the statistics can be set to None. In the code there are 3 separate places where that can happen, i.e. when it is a region, when it is a subdivision and when the country is not in geography_statistics_by_countries. It would be great if we could add tests for all of those, not just to cover these cases but also to better document when the statistics can be None and why.

annaCPR: Those tests don't have to be at this level necessarily.

jamesgorrie: @annaCPR nice catch and a good reminder of why I like table based tests as it really makes you think about the cases you're testing.

PR #710, opened 2025-09-10
refactor: set defaults on search request body on search request

jamesgorrie:
# Description

I am just working my way through how search works as I have been asked and cannot answer.

I haven't written tests, even though they would be quite easy, as this is likely to change significantly - and they do not currently exist.

Hopefully this is clearer with comments and code, if not happy to close it off.

https://linear.app/climate-policy-radar/issue/APP-1137/investigate-how-search-works


backend-api/app/service/search.py
```diff
@@ -646,10 +639,12 @@ def _process_vespa_search_response_families(
     Note: this function requires that results from the cpr sdk library are grouped
           by family_import_id.
     """
+    # we apply a separate
```
katybaulch, line 642: question: what do we apply separately?

backend-api/app/service/search.py
```diff
@@ -646,10 +639,12 @@ def _process_vespa_search_response_families(
...
     vespa_families_to_process = vespa_families[offset : limit + offset]
     all_response_family_ids = [vf.id for vf in vespa_families_to_process]
 
     # TODO: Potential disparity between what's in postgres and vespa
+    # 1. we lookup all
```
katybaulch, line 647:
```suggestion
    # 1. Lookup all families & their metadata
```

backend-api/app/service/search.py
```diff
@@ -850,32 +845,35 @@ def create_vespa_search_params(
...
-        return SearchType.standard
+def set_all_results_default_on_search_request_body(
+    search_request_body: SearchRequestBody,
+):
+    # use the defaults set in the SDK
+    all_results = search_request_body.all_results
 
+    # if we are not text searching we can get all the results
+    # this includes if we have filters applied
+    # otherwise we are limited to the first 500 relevant results
```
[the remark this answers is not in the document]
jamesgorrie: Like me right now. Why is it 500?

jamesgorrie:
I am going to close this as not massively useful

PR #722, opened 2025-09-17
Fix/make searches fast again

henry-cpr:
# Description

The functionality that processes vespa search results in the search endpoint codepath was pretty tangled and slow. 

I have
- Pulled out a bunch of logic into more atomic functions, to improve readability 
- Improved the caching logic, fixing a bug where in-function lookup tables were being recreated repeatedly
- Moved the logic that defines sort-key data to O(1) runtime on model creation, rather than running every  sort-comparator-function call (O(N^2)) 
- Moved the passage-sorting logic outside of the parsing loop, so it runs once rather than on every insert

## Type of change

- [x] Bug fix
- [x] Refactor legacy code

## How Has This Been Tested?

Local tests pass. Haven't been able to smoke test on running system, keen to do that on staging. 

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/backend-api/app/models/search.py
+++ b/backend-api/app/models/search.py
@@ -151,6 +151,7 @@ class SearchResponseDocumentPassage(BaseModel):
     text_block_page: Optional[int] = None
     text_block_coords: Optional[Sequence[Coord]] = None
     concepts: Optional[Sequence[Concept]] = None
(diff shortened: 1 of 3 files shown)


katybaulch (approved):
all seems to make sense to me and can't see anything wrong with the refactor. will let you test on staging - great work!!

PR #739, opened 2025-09-25
fix: add tests for family ordering

jamesgorrie:
# Description
- adds ordering to the families endpoint as we use `LIMIT` which can lead to unpredictable results

> When using LIMIT, it is important to use an ORDER BY clause that constrains the result rows into a unique order. Otherwise you will get an unpredictable subset of the query's rows. You might be asking for the tenth through twentieth rows, but tenth through twentieth in what ordering? The ordering is unknown, unless you specified ORDER BY.

-- [postgres docs](https://www.postgresql.org/docs/current/queries-limit.html)

[This was discovered while paginating through the API looking for deletions](https://github.com/climatepolicyradar/litigation-data-mapper/pull/93).

### DIFF
--- a/families-api/app/router.py
+++ b/families-api/app/router.py
@@ -6,7 +6,7 @@ from typing import Generic, TypeVar
 from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel
 from sqlalchemy import text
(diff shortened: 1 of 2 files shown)


odrakes-cpr (approved):
lgtm

PR #747, opened 2025-10-02
Feature/app 1264 update mcf data download columns

katybaulch:
# Description

We were hardcoding CCLW corpus specific fields to all of our data download files. This PR splits the MCF data download into its own SQL file for maintainability and driveby updates the excel engine used in the CCC data download logic as it is more forgiving of special characters & encodings, and prevents us needing a separate library (which my previous implementation required).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/backend-api/app/repository/download.py
+++ b/backend-api/app/repository/download.py
@@ -32,6 +32,8 @@ def get_whole_database_dump(
     """
     if theme and theme.upper() == "CCC":
         filename = "ccc-download.sql"
+    elif theme and theme.upper() == "MCF":
+        filename = "mcf-download.sql"
     else:
         filename = "download.sql"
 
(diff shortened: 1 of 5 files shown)


henry-cpr (approved):
I think we all know there's a fair amount of fragility in here but I really don't think it's wroth the time re-engineering properly with that SQL! I think there could be errors depending on data quality -- eg if fields are missing or contain unexpected characters. So with that caveat lgtm but if we see errors we should probably pull back and rethink approach! 

oh and do we need to add xlsxwriter to pyproject.toml?

PR #749, opened 2025-10-07
fix: make geographies type filter additive

jamesgorrie:
# Description

Make the type filter additive so
- `type=region` = `regions`
- `type=region&type=subdivision` = `regions AND subdivision`

### DIFF
--- a/geographies-api/app/router.py
+++ b/geographies-api/app/router.py
@@ -216,11 +216,11 @@ async def read_geographies(
         result = geographies.regions + geographies.countries + geographies.subdivisions
     else:
         if GeographyType.region in type:
-            result = geographies.regions
-        elif GeographyType.country in type:
-            result = geographies.countries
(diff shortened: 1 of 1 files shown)


geographies-api/app/router.py
```diff
@@ -216,11 +216,11 @@ async def read_geographies(
...
-            result = geographies.regions
-        elif GeographyType.country in type:
-            result = geographies.countries
-        elif GeographyType.subdivision in type:
-            result = geographies.subdivisions
+            result.extend(geographies.regions)
+        if GeographyType.country in type:
+            result.extend(geographies.countries)
+        if GeographyType.subdivision in type:
+            result.extend(geographies.subdivisions)
```
annaCPR, line 223: How has this not broken any tests? Feels like a fundamental logic change unless I'm missing something.

PR #769, opened 2025-10-27
feat: tag deploys with git sha

jamesgorrie:
# Description
- builds with the tag of the `GITHUB_SHA`
- pushes that along with the `{{tag}}`

This means we can see which SHA corresponds to which named container e.g. `latest`.

👀 
Screenshot 2025-10-27 at 09 21 42


katybaulch (approved):
lgtm!!!

PR #783, opened 2025-10-30
Create End to End ETL Pipeline Flow

odrakes-cpr:
# Description

Refactor the ETL pipeline flow to separate the tasks into well tasks. 

Add cache task which for now uploads the source doc to aws.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


data-in-pipeline/app/util.py
```diff
@@ -1,8 +1,57 @@
...
+    try:
+        if content_type:
+            client.put_object(
+                Body=json_content, Bucket=bucket, Key=key, ContentType=content_type
+            )
+        else:
+            client.put_object(Body=json_content, Bucket=bucket, Key=key)
+
+    except Exception:
+        _LOGGER.exception(f"Uploading {key} encountered an error")
```
jamesgorrie: Do we not want to throw the error here? This will just swallow it.

data-in-pipeline/app/flow.py
```diff
@@ -4,22 +4,56 @@
...
+    """Load document to RDS."""
     load_rds(document)
+    return document
+
 
+@task(log_prints=True)
+def document_pipeline(id: str):
+    """Process a single document through the pipeline."""
+    navigator_document = extract_document(id)
+    cache_document(navigator_document)
```
jamesgorrie: my understanding is that this PR is meant to be implementing the `load` step - is this right?

or are we creating caches between the steps?

I think we should go for the former, and then think of a caching strategy, but what you have here feels like it would work nicely.

data-in-pipeline/tests/test_unit.py
```diff
@@ -1,9 +1,42 @@
...
+        bucket="cpr-production-document-cache",
+        key="navigator/test-123.json",
+    )
+
+
+@patch("app.flow.upload_to_s3")
+def test_cache_document_handles_upload_failure(mock_upload):
+    """Test handling of S3 upload failure."""
+
+    mock_navigator_doc = Mock()
```
katybaulch: question: why Mock over MagicMock, which implements the dunder methods for us too?

data-in-pipeline/app/util.py
```diff
@@ -1,8 +1,57 @@
...
+    :param [str | None] key: filename of the resulting file on s3. Should include
+        the file extension. If not provided, the name of the local file is used.
+    :param [str | None] content_type: optional content-type of the file
+    """
+
+    # Upload the file
+    try:
+        if content_type:
+            client.put_object(
+                Body=json_content, Bucket=bucket, Key=key, ContentType=content_type
```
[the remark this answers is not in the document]
odrakes-cpr: hmmm i think if its not set its stored as binary/octet-stream. I have removed it anyway because we have a default value it will never get to that side of the conditional.

data-in-pipeline/app/flow.py
jamesgorrie: @odrakes-cpr question - why are we caching at this point?

My understanding of this PR is to have the load step in ETL?

i.e. after the transformer load the transformed document into S3.

I am still unclear as to what this cache step is?

PR #788, opened 2025-11-04
fix: tighter typing, naming and new navigator_family pipeline

jamesgorrie:
- adds a new `navigator_family` pipeline
- simplifies the naming a little 
- adds stronger types to the pipeline to ensure we're processing the right type of thing

This is in aid of trying to help come up with a strategy for identifying our documents. Having 2 sources is closer to the reality of where the system will be, and having multiple is always a little trickier than 1.

https://github.com/climatepolicyradar/navigator-backend/blob/b49a8b196f28a7aad20b13515d51fe46a22a21d4/data-in-pipeline/README.md
Screenshot 2025-11-05 at 08 54 42

### DIFF
--- a/data-in-pipeline/README.md
+++ b/data-in-pipeline/README.md
@@ -5,7 +5,8 @@
 ```mermaid
 
 flowchart LR
-    SourceSystem -- SourceDocument --> Identifier -- IdentifiedSourceDocument --> Transformer -- Document --> DocumentStore
+    SourceSystemA -- Extracted --> IdentifierA -- Identified --> Transformer -- Document --> DocumentStore
+    SourceSystemB -- Extracted --> IdentifierB -- Identified --> Transformer
 
 ```
 


data-in-pipeline/app/extract/navigator_family.py
```diff
@@ -0,0 +1,39 @@
...
+
+def extract_navigator_family(import_id: str) -> Extracted[NavigatorFamily]:
+    print(f"Fetching {import_id} from {API_BASE_URL}/families/{import_id}")
+    data = _fetch(import_id)
+    extacted = Extracted(data=data, source="navigator_family")
+    return extacted
+
+
+def _fetch(id: str) -> NavigatorFamily:
+    """Fetch a document from the API.
```
jamesgorrie:
```suggestion
    """Fetch a family from the API.
```

PR #794, opened 2025-11-06
fix: pr labeller uses github API to get changed files

jamesgorrie:
# Description

Removes using git to get the diff as [this is failing](https://github.com/climatepolicyradar/navigator-backend/actions/runs/19141990104/job/54711713562?pr=792) when the head is not in the depth=1 of the origin. 

Instead we use the GitHub API


katybaulch (approved):
lgtm

PR #797, opened 2025-11-10
fix: very basic transformation pattern

jamesgorrie:
# Description

Adds a very basic transformation pattern for the case
"A family has 1 document with a matching title".

I'd like to start getting this running through the pipeline, and then we can start adding more reporting e.g. "We have transformed 9000/50000 documents".

I don't imagine this being the final shape, but definitely a v1 using results.


odrakes-cpr (approved):
vgood vnice

PR #801, opened 2025-11-10
Feature/app 1411 get otel prefect and python loggers to emit the same things

katybaulch:
get otel prefect and python loggers to emit the same things


katybaulch:
> @katybaulch - have you had a peek at [this](https://github.com/PrefectHQ/prefect/issues/9271#issuecomment-2842013922)?
> 
> Maybe we need to install with `otel` extras? `uv add prefect[otel]`

I have - it was linked in one of my first PRs for adding the observability :) I'm not sure we need it as those libraries are meant to be baked into the Docker image already, but if things still aren't working I'll look to add it, thanks!

data-in-pipeline/app/dev.py
katybaulch: Having Sabin as the default breaks `just dev` as that doc isn't available on staging

odrakes-cpr (approved):
looks good

PR #805, opened 2025-11-11
Feature/app 1411 Forward Prefect & Python logs to OTel

katybaulch:
# Description

Prefect has OTel baked in. But that didn't end up meaning what I thought it did. It means the libraries are there and it'll automatially wrap spans for you without you needing to manually observe stuff, but what it _doesn't_ mean is that it automatically forwards the logs from Prefect to OTel - at least not in the current state I have it in with the Dockerfile.

Quick MVP to test my hypothesis, that uses the service manifest file. Next steps are to refactor the /api repo to have a base OTel class that we can inherit from for FastAPI and another for Prefect, to separate the concerns and make sure we are instrumenting them differently. This is obvs a bigger job, so this is an intermediary PR to help make sure we are still capturing logs.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/.trunk/configs/ruff.toml
+++ b/.trunk/configs/ruff.toml
@@ -1,3 +1,5 @@
+target-version = "py312"
+
 # Generic, formatter-friendly config.
 select = ["D3", "E", "F", "TID252", "UP006"]
 


data-in-pipeline/app/logging_handlers.py
```diff
@@ -0,0 +1,204 @@
+import json
+import logging
+import os
+from pathlib import Path
+from threading import Lock
+from typing import Dict, Optional
```
jamesgorrie: sideline: [I tried adding `UP006`](https://docs.astral.sh/ruff/rules/non-pep585-annotation/) - but can only get it to run via `ruff` rather than `trunk.ruff`.

data-in-pipeline/app/logging_config.py
```diff
@@ -8,13 +8,76 @@
...
-        raise RuntimeError("OpenTelemetry logging instrumentor failed") from exc
-else:
-    _LOGGER.debug("OpenTelemetry logging instrumentor disabled")
+        _LOGGER.exception(
+            "Failed to enable OpenTelemetry logging instrumentor.",
+        )
+        raise RuntimeError("OpenTelemetry logging instrumentor failed.") from exc
+
+
+def ensure_logging_active(force_instrumentation: bool = False) -> None:
```
jamesgorrie: suggestion: We seem to always set `force_instrumentation=True` - so this doesn't seem like a [sensible default](https://www.thoughtworks.com/en-gb/insights/topic/sensible-defaults). 

It also means we could probably just remove it as configuration as it just adds complexity for future travels to decide whether they need to use it or not.

data-in-pipeline/app/logging_handlers.py
katybaulch: did you add the rule to the ruff.toml file? I have done this now.

data-in-pipeline/app/logging_config.py
katybaulch: have changed the default to True, but not removing the config as my next PR will wipe this code

jamesgorrie (approved):

.trunk/configs/ruff.toml
```diff
@@ -1,5 +1,5 @@
 # Generic, formatter-friendly config.
-select = ["D3", "E", "F", "TID252"]
+select = ["D3", "E", "F", "TID252", "UP006"]
```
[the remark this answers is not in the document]
katybaulch: Yep - you were missing the `target-version`

PR #814, opened 2025-11-13
Split core OTEL exporter from fastapi specific implementation

katybaulch:
# Description

Instead of having a single OTEL exporter which has generic & fastapi specific code needed for the otel exporters, split them into a generic and specialised class. This is key for the next PR where we introduce the OTEL prefect exporter, which we didn't want to instantiate fastapi specific stuff when it didn't need to.

This is just a refactor of the existing code Henry has written.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [x] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/api/src/api/base_telemetry.py
+++ b/api/src/api/base_telemetry.py
@@ -1,5 +1,10 @@
 """
-Base OpenTelemetry wiring shared across Navigator services.
+Telemetry primitives shared across Navigator services.
+
+This module defines `BaseTelemetry`, the abstraction that bootstraps
+OpenTelemetry clients for downstream frameworks. It establishes tracing,
+structured logging, and exception enrichment that concrete instrumenters
+extend, ensuring each service inherits a consistent baseline configuration.
 """
 
 import logging
--- a/api/src/api/fastapi_telemetry.py
+++ b/api/src/api/fastapi_telemetry.py
@@ -1,4 +1,12 @@
-"""FastAPI-specific OpenTelemetry wiring."""
+"""
+FastAPI integration for Navigator telemetry.
+
+This module defines `FastAPITelemetry`, the first concrete extension of
+`BaseTelemetry`. It adapts the shared bootstrap for FastAPI runtimes by
+binding instrumentation, enriching async exception handling, and 
+wrapping routes so we obtain uniform OpenTelemetry behaviour with
+minimal framework specific code.
+"""
 
 import asyncio
 import logging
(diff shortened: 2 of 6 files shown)


api/src/api/telemetry_base.py
jamesgorrie: nit / question: can we keep class & filenames the same? i.e. `base_telemetry`? This question applies to all the files. It's a tiny, but annoying bit of cognitive load when importing things.

jamesgorrie (commented):
I am battling to see what is happening here - but that's probably just because it is a large refactor and unfamiliarity with the code.

Are we:
- generating a BaseTelemetry class that we expect all OTel instrumentation to extend / inherit from?
- the first instance we have is `FastAPITelemetry`?

If this is right - could we use this as a chance to describe what the abstraction is actually doing, and describe how it's useful?

100% sure it is, but it is not massively evident to me as to how - but I'm assuming that is a knowledge gap that maybe some docs or comments might help with

api/src/api/base_telemetry.py
katybaulch: done, everything is also exported in the init so you can just do `from api import BaseTelemetry, FastAPITelemetry etc`

katybaulch:
> I am battling to see what is happening here - but that's probably just because it is a large refactor and unfamiliarity with the code.
> 
> Are we:
> 
> * generating a BaseTelemetry class that we expect all OTel instrumentation to extend / inherit from?
> * the first instance we have is `FastAPITelemetry`?
> 
> If this is right - could we use this as a chance to describe what the abstraction is actually doing, and describe how it's useful?
> 
> 100% sure it is, but it is not massively evident to me as to how - but I'm assuming that is a knowledge gap that maybe some docs or comments might help with

You got it! We have extracted the shared OpenTelemetry bootstrap that Henry wrote into a BaseTelemetry class, which handles the boring but essential and generic bits of the OTel setup: tracer/logger initialisation, OTLP wiring & exporting, and span enrichment. Framework layers, of which FastAPI is the first, now subclass BaseTelemetry to bolt on their specific concerns (async exception hooks, route wrapping, prefect logger hijacking etc.) whilst inheriting the consistent base config. I have added module level docs to explain that shape so the abstraction is clearer

PR #815, opened 2025-11-13
Add Prefect - OTel - Python logging

katybaulch:
# Description

So - turns out I was right and its entirely possible to configure Prefect to run using environment variables. Problem was, I was baking them into the `data-in-pipeline` Prefect image, but aren't passed to the deployment/run, as we're always running bare `uv run python -m app.deploy` via the just commands rather than running those just commands _against_ the container with the baked in vars. Also because the worker pool doesn't have those env vars baked in either.

When we ran locally and did `just dev` I had OTel traces but not when we ran on the cloud or locally deploy to the cloud e.g., by doing `just deploy data-in-pipeline production` latest followed by `just deploy-to-prefect-from-local` (if circumventing CI) or `just deploy-to-prefect` if in CI.

I can't seem to find where in infra the worker pool default job variables are stored and as that worker pool is shared across all of the platform teams jobs too, I've decided not to overwrite or modify the worker pool job variables themselves.

Instead, I have updated our deploy.py file to extend the default worker pool variables with our OTEL environment variables, and also driveby fixed a bug where runs triggered in the cloud were always calling the staging API rather than the prod API.

The environment variables alone though aren't enough to forward OTel logs to our server - so we need to write a custom exporter to do this. This is contained in the new PrefectTelemetry class.

Other useful context [here](https://github.com/climatepolicyradar/navigator-backend/pull/814#issuecomment-3528057857) and in #814 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/data-in-pipeline/app/navigator_document_etl_pipeline.py
+++ b/data-in-pipeline/app/navigator_document_etl_pipeline.py
@@ -79,11 +79,11 @@ def etl_pipeline(
     id: str,
 ) -> Result[Document, Exception]:
     """ETL pipeline"""
-    logger = get_logger()
-    logger.info("ETL pipeline started")
+    _LOGGER = get_logger()
+    _LOGGER.info("ETL pipeline started")
     extracted_result = extract(id)
     if not is_successful(extracted_result):
-        logger.error(f"Extraction failed for {id}: {extracted_result.failure()}")
+        _LOGGER.error(f"Extraction failed for {id}: {extracted_result.failure()}")
         return Failure(extracted_result.failure())
     extracted = extracted_result.unwrap()
     identified = identify(extracted)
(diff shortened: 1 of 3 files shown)


data-in-pipeline/app/navigator_document_etl_pipeline.py
```diff
@@ -25,10 +22,13 @@ def extract(
     document_id: str,
 ) -> Result[ExtractedEnvelope[NavigatorDocument], Exception]:
     """Extract"""
+    _LOGGER = get_logger()
```
jamesgorrie: question: why have we moved this to be initialised at each step?

data-in-pipeline/Dockerfile
katybaulch: See PR description - the env vars are right BUT including them in the dockerfile is the wrong move, they needed to be in the Prefect job variables

PR #828, opened 2025-11-27
fix(DIP): transformer refactor and simplification

jamesgorrie:
# Description
- move to having 1 transformer
- for each `NavigatorFamily` => 1 `Document`
- for each `NavigatorDocument` => 1 `Document`
- in some cases, a `NavigatorDocument` `is_version_of` a `NavigatorFamily` - this will allow us to apply merging logic later, and ensure we don't surface them in search
- If a document is not `is_version_of` it is considered a `member_of`

This makes the logic way more easy to understand, and creates a 1-1 relationship between families and document to the new `Document` model.


jamesgorrie:
> To be honest I preferred the previous pattern of the transformer logic

I agree if we were sticking to the paradigm of having the family being the branching point of the document. 

So it is not the logic that I am changing - but rather the paradigm. We will have 1 `Document` / `NavigatorFamily` and 1 `Document` / `NavigatorDocument`.

Each document and family will mostly be transformed the same. 

@annaCPR is it the outputs (i.e. the test data) that you do not like, or just the implementation of how we get there?

PR #836, opened 2025-12-02
Make the aurora cluster identifier more meaningful

annaCPR:
# Description
Set `cluster_identifier` when creating the Aurora cluster so that it has a more meaningful name and reflects the AWS environment.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/data-in-pipeline/infra/__main__.py
+++ b/data-in-pipeline/infra/__main__.py
@@ -160,7 +160,8 @@ aurora_subnet_group = aws.rds.SubnetGroup(
 
 
 aurora_cluster = aws.rds.Cluster(
-    f"{name}-aurora-cluster",
+    f"{name}-{environment}-aurora-cluster",
+    cluster_identifier=f"{name}-{environment}-aurora-cluster",
     engine="aurora-postgresql",
     engine_version="17.6",
     master_username=credentials.apply(lambda creds: creds["username"]),


data-in-pipeline/infra/__main__.py
```diff
@@ -160,7 +160,8 @@ def generate_secret_key(project: str, aws_service: str, name: str):
 
 
 aurora_cluster = aws.rds.Cluster(
-    f"{name}-aurora-cluster",
+    f"{name}-{environment}-aurora-cluster",
+    cluster_identifier=f"{name}-{environment}-aurora-cluster",
```
jamesgorrie: I am always a little wary of putting the `environment` into the name? 

Reason being is that the environment is forked right at the top of the chain i.e. the AWS account, so not sure what value it adds and means you need to then use this pattern downstream making all apps etc environment aware.

For example, the AppRunner instance could just connect to `data-in-pipeline-aurora-cluster` without needing any config.

As soon as this differs from env to env, we then need to pass that value down and make the service configurable, with not much benefit.

PR #849, opened 2025-12-08
Feature/app 1506 update bastion to be able to connect to aurora

katybaulch:
# Description

Allow our bastion to connect to Aurora so we can remote into it.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Refactor legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/data-in-pipeline/infra/__main__.py
+++ b/data-in-pipeline/infra/__main__.py
@@ -71,6 +71,7 @@ aurora_security_group = aws.ec2.SecurityGroup(
         ),
     ],
     tags=tags,
+    opts=pulumi.ResourceOptions(depends_on=[aws_env_stack]),
 )
 
 eu_west_1a_private_subnet_id = aws_env_stack.get_output("eu_west_1a_private_subnet_id")


data-in-pipeline/infra/__main__.py
```diff
@@ -117,13 +173,38 @@
...
+)
+pulumi.export(
+    f"{name}-{environment}-aurora-instance-ids",
+    [instance.identifier for instance in aurora_instances],
+)
+pulumi.export(f"{name}-{environment}-aurora-endpoint", aurora_cluster.endpoint)
+pulumi.export(
+    f"{name}-{environment}-aurora-reader-endpoint", aurora_cluster.reader_endpoint
+)
+
```
jamesgorrie: ☝️ @katybaulch are we planning on using these anywhere? I do not think we will as most of these values will be used in this stack.

If not can we remove them as we might create inadvertent (over explicit) couplings. 

I have always seen exports as the equivalent to API contracts. We do not currently have a way of letting know any upstream services these values have changed, so if they do couple to them we can potentially break those services, so it might be best to avoid them till them.

We have seen this with the bastion updating which has broken other services in our stack.

PR #850, opened 2025-12-09
fix(dip): remove depends_on from pulumi resources

jamesgorrie:
# Description

I was reviewing this PR - and got interested in the dependency resolution from Pulumi as it is usually one of the great things about IoC libraries (Terraform I was familiar with, which powers a lot of Pulumi).

Turns out Pulumi does a lot of the heavy lifting for us, so we do not need to define the dependencies except in some circumstances (I am not sure which from the documentation).

> Pulumi automatically tracks dependencies between resources when you supply an input argument that came from another resource’s output properties.

-- [Pulumi docs](https://www.pulumi.com/docs/iac/concepts/resources/options/dependson/)

☝️ this covers our bases


jamesgorrie:
@katybaulch 

>  the few arbitrary cases in which it is required

It is not arbitrary - it always infers if an output of a resource is the input of another, that they are dependant. The docs are quite good at describing this. I think there is also internal dependency trees around how AWS in general is provisioned in the underlaying Terraform libs.

It is when you make arbitrary couplings, you need to let Pulumi know which is the intended use of `depends_on`.

To complete citing the docs

> you may need to explicitly specify additional dependencies that Pulumi doesn’t know about but must still respect. This might happen if a dependency is external to the infrastructure itself—such as an application dependency—or is implied due to an ordering or eventual consistency requirement.

---

>  I have been burnt by this before 

I would be super interested to here what the circumstances were as I am not sure which error case this implementation would catch?

A large benefit of using IoC is the dependency chain being built for you. 

I completely agree that, on average, explicit it generally better, mainly when things resolve to the equivalent of `Any` and so you have no insight into what is happening. However - the Pulumi dependency tree is highly managed by its internal state - and I would prefer to rely on it.

katybaulch:
> @katybaulch
> 
> > the few arbitrary cases in which it is required
> 
> It is not arbitrary - it always infers if an output of a resource is the input of another, that they are dependant. The docs are quite good at describing this. I think there is also internal dependency trees around how AWS in general is provisioned in the underlaying Terraform libs.
> 
> It is when you make arbitrary couplings, you need to let Pulumi know which is the intended use of `depends_on`.
> 
> To complete citing the docs
> 
> > you may need to explicitly specify additional dependencies that Pulumi doesn’t know about but must still respect. This might happen if a dependency is external to the infrastructure itself—such as an application dependency—or is implied due to an ordering or eventual consistency requirement.
> 
> > I have been burnt by this before
> 
> I would be super interested to here what the circumstances were as I am not sure which error case this implementation would catch?
> 
> A large benefit of using IoC is the dependency chain being built for you.
> 
> I completely agree that, on average, explicit it generally better, mainly when things resolve to the equivalent of `Any` and so you have no insight into what is happening. However - the Pulumi dependency tree is highly managed by its internal state - and I would prefer to rely on it.

My comment around 'arbitrary cases' referred to your 'except in some circumstances (I am not sure which from the documentation).', so thanks for explaining further, that is clear! 

I can't remember the specifics of whatever I was talking about here but when I do I'll come back and add them 😂 . I'm happy with what you've said and my initial comment was definitely not aimed at blocking your PR from being merged - I understand they aren't explicitly needed:)

PR #886, opened 2026-01-05
fix: add otel to dip dependencies

jamesgorrie:
# Description

- adds otel to the dip dependencies

We've piggy backing on this being available at the root due to other workspaces needing it. \

[This has started breaking when we are trying to run some upgrades](https://github.com/climatepolicyradar/navigator-backend/actions/runs/20660716209/job/59322473914?pr=885), so this fixes that.


katybaulch (approved):
nice temp fix :)

PR #891, opened 2026-01-06
Refinements to merge-to-main workflow

jamesgorrie:
# Description

- introduces `deploy-override` to the base `justfile` which any extending services can use (follows the pattern of [`prebuild`](https://github.com/climatepolicyradar/navigator-backend/blob/55be61700da884aa60e10f71a13fcf815aad9187/justfile#L43-L47))
- ☝️ same for `test-override`
- uses the same deployment for all services including `data-in-pipeline`
- because of ☝️ simplifies the `merge-to-main` workflow
- removes `cd-microservices-production` as this seems to be a replica of this workflow

I'll be moving to the `pull-request` workflow after this

👀 [An example of a successful run](https://github.com/climatepolicyradar/navigator-backend/actions/runs/20745701280) 👀

### DIFF
--- a/.github/workflows/merge-to-main.yml
+++ b/.github/workflows/merge-to-main.yml
@@ -27,21 +27,25 @@ jobs:
               - 'families-api/**'
               - 'api/**'
               - 'uv.lock'
+              - pyproject.toml
 
             geographies-api:
               - 'geographies-api/**'
               - 'api/**'
               - 'uv.lock'
+              - pyproject.toml
 
             concepts-api:
               - 'concepts-api/**'
               - 'api/**'
               - 'uv.lock'
+              - pyproject.toml
 
             data-in-pipeline:
               - 'data-in-pipeline/**'
               - 'api/**'
               - 'uv.lock'
+              - pyproject.toml
 
   test:
     needs: changes


.github/workflows/merge-to-main.yml
```diff
@@ -13,78 +12,61 @@ permissions:
   id-token: write
   contents: read
 jobs:
-  determine-changed-services:
+  changes:
```
katybaulch: question: why rename the job - the previous was more descriptive?

.github/workflows/merge-to-main.yml
```diff
@@ -13,78 +12,61 @@ permissions:
...
           filters: |
             families-api:
               - 'families-api/**'
               - 'api/**'
+              - 'uv.lock'
 
             geographies-api:
               - 'geographies-api/**'
               - 'api/**'
+              - 'uv.lock'
```
katybaulch: are we happy with only uv.lock?? i would also add the pyproject.toml -- sometimes people forget to uv sync and i think this could get us into a sticky sitch

.github/workflows/merge-to-main.yml
```diff
@@ -116,55 +97,31 @@ jobs:
...
-          if [ "${{ matrix.environment }}" = "staging" ]; then
-            echo "webhook_url=${{ secrets.SLACK_WEBHOOK_URL_STAGING_UPDATES_CHANNEL }}" >> $GITHUB_OUTPUT
-          elif [ "${{ matrix.environment }}" = "production" ]; then
-            echo "webhook_url=${{ secrets.SLACK_WEBHOOK_URL_PRODUCTION_UPDATES_CHANNEL }}" >> $GITHUB_OUTPUT
-          fi
-
       - uses: slackapi/slack-github-action@v2.1.1
         with:
-          webhook: ${{ steps.webhook_url.outputs.webhook_url }}
+          webhook: ${{ secrets.ENV_UPDATES_CHANNEL_SLACK_WEBHOOK_URL }}
```
katybaulch: praise: nice!!

jamesgorrie: note to self: 🧹 remove the sunsetted secrets.

.github/workflows/merge-to-main.yml
```diff
@@ -13,78 +12,61 @@ permissions:
...
       - uses: actions/checkout@v6
       - uses: dorny/paths-filter@v3
-        id: changes
+        id: filter
         with:
           filters: |
             families-api:
               - 'families-api/**'
               - 'api/**'
+              - 'uv.lock'
```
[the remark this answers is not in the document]
katybaulch: Could we do both? My motivations being it doesn't hurt to guard against them being out of sync but also make sure we have a reliable fallback that will trigger on changes to pyproject.toml too?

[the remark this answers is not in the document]
katybaulch: I've just had a little look and it looks like the version of Trunk we are using has some built in [`uv` actions](https://github.com/trunk-io/plugins/tree/main/actions/uv) we can use at precommit/prepush level, which means we might not need to write one ourselves & it can form a standard part of the CI checks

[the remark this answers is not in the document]
katybaulch: Specifically, the `uv-lock` action looks to be what we're after

.github/workflows/merge-to-main.yml
katybaulch: The point I am trying rather ineloquently to make is that no where in the docs for the action you are using in that step (that you linked above), does it specify that the name of the job using the action should be 'changes'. That is just the name that the author decided to use for the job in their examples.

Because of this, I don't understand why would we change the name of the job to be less meaningful that what was already there? Perhaps I am too big of a fan of descriptive naming, as 'changes' on its own I do not find particularly meaningful. 

Just to be clear - this isn't blocking and I'm not trying to be obstinate. I'm just not understanding the rationale you've given 😅

PR #896, opened 2026-01-06
fix: update pull-request workflow

jamesgorrie:
# Description

- matches `pull-request` => `merge-to-main` `changes` step
- matches `pull-request` => `merge-to-main` `deploy` step but in the `staging` environment and on label `deploy:staging`
- adds code-quality (Trunk.io) to pull-request workflow

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -132,9 +132,9 @@ jobs:
           if [ -n "${{ steps.pr_data.outputs.pr_link }}" ]; then
             PR_URL="https://github.com/${{ github.repository }}/pull/${{ steps.pr_data.outputs.pr_number }}"
             PR_NUM="${{ steps.pr_data.outputs.pr_number }}"
-            MESSAGE="🚀 Deploying <$PR_URL|PR #$PR_NUM> (<$COMMIT_URL|$COMMIT_SHORT>) *${{ matrix.service }}* to ${{ matrix.environment }}"
+            MESSAGE="🚀 Deploying feature-branch <$PR_URL|PR #$PR_NUM> (<$COMMIT_URL|$COMMIT_SHORT>) *${{ matrix.service }}* to ${{ matrix.environment }}"
           else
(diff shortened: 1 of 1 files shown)


katybaulch (changes requested):
all good aside from the new data-in-pipeline-load-api lines being removed - these need adding back in please 🙏🏻

.github/workflows/pull-request.yml
```diff
@@ -17,118 +17,140 @@ permissions:
...
       # we use python / uv in some build scripts and testing
       - uses: actions/setup-python@v6
         with:
           python-version: "3.12"
+
       - uses: astral-sh/setup-uv@v7
       - run: uv sync --all-packages
-      - run: just deploy ${{ matrix.service }} ${{ matrix.environment }} latest
+
+      - id: pr_data
```
katybaulch: polish: the code here seems to be copy pasted from the deploy actions from merge to main which is okay, but it's more complex than it needs to be. as we are inside a PR, we can get the PR number directly from the PR context rather than having to run extra queries. this entire step is not needed

.github/workflows/pull-request.yml
```diff
@@ -17,118 +17,140 @@ permissions:
...
+            echo "pr_number=$PR_NUMBER" >> $GITHUB_OUTPUT
+            echo "pr_link=https://github.com/${{ github.repository }}/pull/$PR_NUMBER" >> $GITHUB_OUTPUT
+          fi
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - run: |
+          COMMIT_URL="https://github.com/${{ github.repository }}/commit/${{ github.sha }}"
+          COMMIT_SHORT=$(echo "${{ github.sha }}" | cut -c1-7)
+          if [ -n "${{ steps.pr_data.outputs.pr_link }}" ]; then
```
katybaulch: polish: we will always have a PR link as this GHA runs on pull_request context - i think it is `github.event.pull_request.html_url` or similar?

PR #915, opened 2026-01-08
fix: few fixes for load-api connections

jamesgorrie:
# Description
- parse username, password from password manager
- use `sslmode=require`
- add lambda `requirements.txt`

The main issue here was that the value from secret store is a JSON string `{"username": "xxx", "password": "xxx"}` - we were passing that whole string as the password.

There is a follow up infra PR coming soon.

Follow up from https://github.com/climatepolicyradar/navigator-backend/pull/910

### DIFF
--- a/data-in-pipeline-load-api/app/settings.py
+++ b/data-in-pipeline-load-api/app/settings.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import sys
 
 from pydantic import SecretStr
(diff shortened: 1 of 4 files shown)


data-in-pipeline-load-api/app/settings.py
```diff
@@ -39,9 +38,7 @@ class Settings(BaseSettings):
     # 'prefer' tries SSL but falls back to non-SSL for local dev (validates
     # certs if SSL is used). For production RDS without cert validation,
     # set db_sslmode=require via environment variable.
-    db_sslmode: str = (
-        "prefer" if os.getenv("ENV", "development") != "production" else "require"
-    )
+    db_sslmode: str = "require"
```
katybaulch: this wont work locally:(

PR #954, opened 2026-01-20
feat(dip): data-in-load-api

jamesgorrie:
# Description

A relatively small setup that includes a `Prefect Flow => ECS Task => AppRunner` via private connections in AWS.

Watch this space.

[Successful run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0696f9ab-9c37-75a5-8000-358601796a6a?preview=true&tab=logs)

PR #962, opened 2026-01-21
Secure load api

annaCPR:
# Description

The networking @annaCPR talk about below is the right design and implementation.

The Prefect tasks, when provisioned from Prefect, are not running in the VPC that we have provisioned everything else in, and, strangely, all the other Prefect infrastructure is in 🤷 [^1].

This is the [magic][magic], as tested [here](https://github.com/climatepolicyradar/navigator-backend/pull/956) and [here](https://github.com/climatepolicyradar/navigator-backend/pull/954).

I will be investigating further into the "why?" and "how can we not put ourselves through this again".

[magic]: https://github.com/climatepolicyradar/navigator-backend/pull/962/changes#diff-cdc1ed8c9c06726b8fe88259f4f59b5da16d8a8b5c8e3506e77dea05d2aba5c7R285-R292

---

What we've done:

- added a VpcIngressConnection that was missing to connect the VpcEndpoint with AppRunner
- used the AppRunner service url as the load api url
- switched back to https
- used cidr blocks instead of prefect security group for the ingress rules of the VPC endpoint security group

This was still giving us timeout errors.

We then:
- removed egress config from AppRunner (to simplify things even further)
- also removed Route53 private hosted zone and A record that was tried before but also did not work
- commented out the body of the health check in load API so that it doesn't call Aurora
- in the DIP flow we added a call to /load/health with a log of the response

Unfortunately connection still times out.

Obviously, throughout this: is_publicly_accessible=False on the AppRunner ingress configuration.


data-in-pipeline-load-api/app/routers.py
```diff
@@ -17,19 +16,20 @@
...
-        raise HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e)
-        )
-
-    if not is_healthy:
-        raise HTTPException(
-            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-            detail="Database connection unhealthy",
-        )
+    # FIXME: https://github.com/climatepolicyradar/navigator-backend/issues/963
```
jamesgorrie: https://github.com/climatepolicyradar/navigator-backend/issues/963

data-in-pipeline/infra/__main__.py
```diff
@@ -191,7 +191,8 @@
     preferred_backup_window="02:00-03:00",
     iam_database_authentication_enabled=False,  # TODO: Reenable later
     preferred_maintenance_window="sun:04:00-sun:05:00",
-    deletion_protection=True,
+    # FIXME: https://github.com/climatepolicyradar/navigator-backend/issues/964
```
jamesgorrie: https://github.com/climatepolicyradar/navigator-backend/issues/964

jamesgorrie:
Related to #958

PR #1030, opened 2026-01-29
fix: deduplicate document label relationships

odrakes-cpr:
**Fix: Prevent duplicate label constraint violations during document seeding**

Database seeding was failing with UniqueViolation errors when documents had duplicate labels with the same (label_id, type) combination. This occurred when multiple sources (e.g., corpus_type and category) mapped to the same label value like "Legal case".
Error example:
```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "pk_documentlabellink"
DETAIL: Key (document_id, label_id)=(Sabin.family.103781.0, Legal case) already exists.
```

**Solution**
Added deduplicate_labels() function to remove duplicate DocumentLabelRelationship objects before creating documents. The function:

Deduplicates by (label.id, type) tuple to match the database constraint
Preserves insertion order, keeping the first occurrence
Prevents IntegrityError on the pk_documentlabellink composite primary key

**Changes**

app/transform/navigator_family.py:

Added deduplicate_labels() utility function
Applied deduplication in _transform_navigator_family() before returning Document

tests/transform/test_navigator_family.py:

Added navigator_family_with_duplicate_legal_case fixture to reproduce the duplicate scenario

### DIFF
--- a/data-in-api/docker-compose.yml
+++ b/data-in-api/docker-compose.yml
@@ -33,6 +33,10 @@ services:
       load_database_url: db
       db_port: 5432
       db_name: data-in-api
+      DB_URL: db
+      DB_NAME: data-in-api
+      DB_USERNAME: data-in-api
+      DB_SECRETS: '{"password": "data-in-api", "username": "data-in-api"}'
     command:
       [
         uv,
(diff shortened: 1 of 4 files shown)


jamesgorrie (approved):
less labels, lovely!

PR #1031, opened 2026-01-29
feat: move data-in-api infra from data-in-pipeline

jamesgorrie:
# Description

- decouples the `data-in-api` and `data-in-pipeline` infrastructure
  - We were using the same security groups etc for both services, which did not follow the pattern of least privilege
- this will now deploy the relevant infra when using the `just deploy-infra` recipe

[Issue](https://linear.app/climate-policy-radar/issue/APP-1652/move-data-in-api-apprunner-=-data-in-apiinfra)


katybaulch (approved):
Nice! Good to get this split out!

PR #1033, opened 2026-02-02
refactor: switch bulk create endpoint to bulk update 

odrakes-cpr:
Switch to bulk update to run full import of navigator documents. 

We already imported some documents as part of a test.


annaCPR (approved):
lgtm

PR #1035, opened 2026-02-02
Update families api request page size to 100 from default of 10 

odrakes-cpr:
Increase the Families API request page size from the default 10 to 100, matching the API’s maximum supported page size.

See below. 
https://github.com/climatepolicyradar/navigator-backend/blob/99ed4b562e16d2e8e45b4b375534d5150d695093/families-api/app/router.py#L56

### DIFF
--- a/.github/workflows/merge-to-main.yml
+++ b/.github/workflows/merge-to-main.yml
@@ -15,7 +15,8 @@ jobs:
   changes:
     runs-on: ubuntu-latest
     outputs:
-      services: ${{ steps.filter.outputs.changes }}
+      services: ${{ steps.parse.outputs.services }}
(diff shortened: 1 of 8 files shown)


data-in-pipeline/app/extract/connectors.py
```diff
@@ -310,7 +310,7 @@ def fetch_all_families(
             with log_context(page_number=page):
                 try:
                     logger.info(f"Fetching families page {page}")
-                    response_json = self.get(f"families/?page={page}")
+                    response_json = self.get(f"families/?page={page}&page_size=100")
```
katybaulch: question: why 100 specifically for the page size? do we want to make this configurable in the function definition with a default value?

PR #1058, opened 2026-02-03
fix: add missing write to document document table

odrakes-cpr:
## Summary

This PR adds first-class support for **document–document relationships** to the persistence layer.
Incoming `DocumentDocumentRelationship` objects are now stored, updated, and removed in the database as part of the existing document upsert flow, with accompanying tests to validate behavior.

---

The API already accepts `DocumentDocumentRelationship` objects (e.g. `has_member`, `member_of`) via the `Document` schema, but these relationships were not previously persisted. As a result:

* Relationships were silently dropped during document upserts
* GET endpoints could never return document–document relationships
* The data model was incomplete relative to the API contract

This change closes that gap and makes document relationships durable and queryable.

---

### Database / persistence

* Added upsert logic for document–document relationships
* Relationships are stored idempotently using `INSERT … ON CONFLICT`
* Removed relationships are deleted to keep DB state in sync with input
* Target documents are ensured to exist before relationships are created

### DIFF
--- a/.github/workflows/merge-to-main.yml
+++ b/.github/workflows/merge-to-main.yml
@@ -42,6 +42,13 @@ jobs:
               - 'uv.lock'
               - pyproject.toml
 
+            data-in-api:
+              - 'data-in-api/**'
+              - 'api/**'
+              - 'data-in-models/**'
+              - 'uv.lock'
+              - pyproject.toml
+
             data-in-pipeline:
               - 'data-in-pipeline/**'
               - 'api/**'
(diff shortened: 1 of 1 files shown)


data-in-pipeline-load-api/tests/test_upsert_documents.py
```diff
@@ -199,3 +221,329 @@ def test_shared_label_does_not_raise_error(session):
...
+#     ).all()
+
+#     assert len(relationships_a) == 1
+#     assert relationships_a[0].related_document_id == "doc-b"
+#     assert relationships_a[0].relationship_type == "parent_of"
+
+#     assert len(relationships_b) == 1
+#     assert relationships_b[0].related_document_id == "doc-a"
+#     assert relationships_b[0].relationship_type == "child_of"
+
```
jamesgorrie: @odrakes-cpr ☝️ what do we want to do with this?

annaCPR (approved):
lgtm

PR #1061, opened 2026-02-03
Create shipyard.yml for local version of all services

katybaulch:
# Description

Needed for Shipyard.

## How Has This Been Tested?

## Reviewer Checklist

- [ ] The code is sufficiently commented, particularly in hard-to-understand
      areas
- [ ] No FIXMEs remain


docker-compose.yml
```diff
@@ -0,0 +1,235 @@
+# Root-level compose for Shipyard: all backend services, no test profiles/containers.
+# Build from repo root: docker compose build
+# Services use internal networking; expose app ports for routing.
+# Ensure backend-api/.env exists (e.g. copy from backend-api/.env.example) for backend + backend_db.
+
+services:
+  # ---------------------------------------------------------------------------
+  # backend-api (Vespa + Postgres + FastAPI)
+  # ---------------------------------------------------------------------------
```
jamesgorrie: suggestion: Given backend-api
- is deprecated
- doesn't interface with anything else (no e2e tests)
- is hard to manage

Should we be building out more infra we would then need to manage?

docker-compose.yml
jamesgorrie: thought: [some interesting reading here](https://docs.docker.com/compose/how-tos/multiple-compose-files/). I have used `docker compose` merging before, and it is quite useful.

docker-compose.yml
katybaulch: It's a pretty key part of the API still - so it needs to be here I'd argue!

PR #1085, opened 2026-02-04
fix(data-in-api): add prefix to API docs

jamesgorrie:
# Description
To make the docs available here https://api.climatepolicyradar.org/data-in/docs


katybaulch (approved):
lgtm

PR #1172, opened 2026-04-09
Feature/app 1957 handle sabin documents that have missing parent concepts

odrakes-cpr:
### Problem

When transforming litigation concepts, if a concept referenced a parent label that didn't exist, the code raised a exception directly, which was silenced so that we could get transformations in. We need a way to gather a list of these transformation errors and relay this back to the Sabin team so that can fix the erroring data.

### Changes

- Updated `_transform_litigation_concepts_to_label_relationships` to return `Result[list[LabelRelationship], CouldNotTransform]` instead of `list[LabelRelationship]`, wrapping the success path in `Success(...)` and the missing parent case in `Failure(...)`
- Updated `_transform_navigator_family` to pattern match on the result and propagate any `Failure` up the call stack
- Updated `_transform_navigator_family`'s return type annotation to the correct two-parameter form `Result[Document, CouldNotTransform]`

### Result

Missing parent labels are now handled as a `Failure` that propagates cleanly through `transform` → `transform_navigator_family` → the pipeline loop, where it is logged and recorded as a metric rather than crashing the pipeline.

### DIFF
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -59,7 +59,7 @@ MCF_ATTRIBUTE_KEYS = {
 
 def transform_navigator_family(
     input: Identified[NavigatorFamily],
-) -> Result[list[Document], NoMatchingTransformations]:
+) -> Result[list[Document], CouldNotTransform | NoMatchingTransformations]:
     logger = get_logger()
     with log_context(import_id=input.id):
         logger.info(f"Transforming family with {len(input.data.documents)} documents")
@@ -68,6 +68,9 @@ def transform_navigator_family(
             case Success(d):
                 logger.info(f"Transform completed, produced {len(d)} documents")
                 return Success(d)
+            case Failure(error):
+                logger.warning(f"Transformation failed: {error}")
+                return Failure(error)
 
         logger.warning("No matching transformation found")
         return Failure(NoMatchingTransformations())
@@ -456,7 +459,6 @@ def _transform_litigation_concepts_to_label_relationships(
     concepts: list[NavigatorConcept],
     family_import_id: str,
 ) -> list[LabelRelationship]:
-    logger = get_logger()
     """
     Convert litigation concepts into label relationships with subconcept hierarchies.
 
@@ -498,11 +500,7 @@ def _transform_litigation_concepts_to_label_relationships(
         for parent_name in concept.subconcept_of_labels:
             parent = label_by_name.get((concept.relation, parent_name))
             if parent is None:
-                # TODO: we should accumulate these errors and report them somewhere.
-                # raise ValueError(
-                #     f"Unknown parent label {parent_name!r} in relation {concept.relation!r}. See family {family_import_id!r} for details."
-                # )
-                logger.error(
+                raise CouldNotTransform(
                     f"Unknown parent label {parent_name!r} in relation {concept.relation!r}. See family {family_import_id!r} for details."
                 )
             else:
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -103,7 +103,11 @@ def transform(
     """
     Transform
     """
-    document_from_family = _transform_navigator_family(input.data)
+    result = _transform_navigator_family(input.data)
+    if isinstance(result, Failure):
+        return result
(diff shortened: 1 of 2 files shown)


data-in-pipeline/app/transform/navigator_family.py
```diff
@@ -858,11 +867,13 @@ def _transform_navigator_family(navigator_family: NavigatorFamily) -> Document:
...
-                navigator_family.concepts, navigator_family.import_id
-            )
-        )
+        match _transform_litigation_concepts_to_label_relationships(
+            navigator_family.concepts, navigator_family.import_id
+        ):
+            case Success(litigation_labels):
+                labels.extend(litigation_labels)
+            case Failure(e):
+                return Failure(e)
```
jamesgorrie, line 876: clarity: so we are saying that if we cannot transform the litigation concepts the whole document is a failed transform.

What happens to this document then? It feels like skipping the whole document is going to land us up in a world of missing data, but if we fail on the labels transform and report (slack?) on that, we would have complete but incorrect data, which feels better? But that's a feely thing - what do you think?

odrakes-cpr: Well yes we are saying that we if we cannot transform the litigation concepts the whole document is a failed transform, which fundamentally is how I think it should work given that is a key part of a litigation document. 

But in reality, all of the litigation documents are already in the database/aurora, so really those documents already show up in the data-in-api, what would happen would be that there would be no legal concepts labels on the family page. I assume we could handle that in the frontend fairly easily - with some sort of fallback component. 

It does mean that fresh and new Sabin documents however will not be transformed if there is incomplete data, and I am of the feeling that is the right way to approach this.

jamesgorrie: So the result is that if we cannot transform the document we will show an old version of the whole document?

I can see this being quite hard to understand when describing it to a user, which is a good benchmark.

And potentially problematic if they want to update e.g. the title urgently.

could we flag the document as `attributes.status=failed_transform` or something that we can then flag up? Maybe with attached errors? My fear is these get swallowed, we forget about this, and have a few days working it, and also lose trust in the partner as we can't describe the problem easily?

odrakes-cpr: This is a good conversation, but I would like to ask what is the point of the `CouldNotTransform` failure type then because to me this seems like the perfect candidate for it and adding an attribute status just goes against that. 

I disagree it will be a problem for a user wanting to update a title because if they’re in the Wordpress account surely they will just update the missing concept. 

I would like to push back on this specifically `So the result is that if we cannot transform the document we will show an old version of the whole document?`  because that's how the system now works. Excusing this specific sabin example if there is a document that can't be transformed or cleansed for whatever reason, say you've added some new transformation functionality, you will see an old version of the whole document. 

We have a [specific slack notification in place for this specific error ](https://github.com/climatepolicyradar/navigator-backend/blob/7a496dbc8e8df84e711a2b07f9c0355533f2a090/data-in-pipeline/app/navigator_family_etl_pipeline.py#L407) so i doubt it would get missed. I personally would prefer if we ran the pipeline gave Sabin this error log and then tell them to fix this data. One benefit is that we are actually getting an insight into the bad data here, which is ironically currently being swallowed by the current `automatic_updates pipeline`.

 This is the first time we have seen this error without going and digging deep into the logs for the `automatic_updates pipeline.` And to push back, the problem can be described very easily as its a very specific niche error which we illustrate well in the logs. @jamesgorrie

annaCPR: Having discussed this, I am leaning towards agreeing with @odrakes-cpr. I think allowing a document through with incorrect or incomplete concepts could lead to FE errors that would potentially live in the UI until the data is corrected by Sabin (plus it might take a while to figure out what the error is exactly depending on how we handle it in the FE code). Whereas if we block the document from being transformed in the first place and surface that error immediately, it can be fixed sooner. And as @odrakes-cpr said, most documents are already transformed. 
@jamesgorrie's point about urgent updates to something like title being blocked by this is a valid one though and we should definitely keep an eye on this. But I'd still rather surface errors early and force a data fix rather than allow incorrect data or potentially errors to flow to the FE.
In either case, we can monitor and change this if needed in the future.

annaCPR (approved):
Let's give this a go

PR #1174, opened 2026-04-13
feat: transform the collection slugs to attributes

odrakes-cpr:
# What does this do?
Per title
## Why is it needed?
The frontend needs this value mapped for the BFF layer
## How was it tested?
Updated tests to account for this new value


annaCPR (approved):
Nicely done!

PR #1175, opened 2026-04-13
Change attributes type

annaCPR:
# What does this do?

- Changes the type of  `attribute` column on `label` and `document` tables in Aurora from SQLAlchemy JSON => PostgresQL JSONB
- Updates the test fixtures so that alembic tests can run inside docker against a test postgres DB instead of SQLite
- Removes the `justfile` from the `data-in-pipeline-load-api` directory as all the tests are now run with a single command 

## Why is it needed?

- As per the table below, JSONB type allows for direct equals comparisons (amongst other things), which will make it a lot easier to compare and update attributes on labels and documents

Screenshot 2026-04-13 at 09 28 25

- SQLite does not support the JSONB type so we could no longer run those tests using pytest locally against the built in engine

## How was it tested?

- Alembic unit tests


odrakes-cpr (commented):
Nice

odrakes-cpr (approved):

PR #1176, opened 2026-04-13
fix: only map label label relationships for labels with allowed types

odrakes-cpr:
# What does this do?
All labels have ids, and labels are not specifically tied to a corpus or document.

Labels can also have relationships with other labels. 

This means that if a Sabin Document and UNFCC document both have the UNFCC label, we attach all UNFCCC label relationships to the documents, even though currently only Sabin documents have that label-label relationships. 

For now we are only mapping that label label relationships to Sabin documents. I have added a note for future visitors. 

UNFCC 

```
{
"type": "deprecated_category",
"value": {
"attributes": {},
"id": "UNFCCC",
"type": "deprecated_category",
"value": "UNFCCC",
"labels": [
{
"type": "subconcept_of",
"value": {
"attributes": {},
"id": "International Law",
"type": "deprecated_category",
"value": "International Law",
"labels": []
},
"timestamp": null
}
]
},
```

Sabin 

```
"type": "legal_concept",
"value": {
"attributes": {
"attribution_url": "https://climate-laws.org/",
"corpus_text": "<p>This document was downloaded from the <a href=\"https://unfccc.int/\" target=\"_blank\"> UNFCCC website</a>. Please check <a href=\"https://unfccc.int/this-site/terms-of-use\" target=\"_blank\"> terms of use </a> for citation and licensing of third party data.</p>",
"corpus_image_url": ""
},
"id": "UNFCCC",
"type": "agent",
"value": "UNFCCC",
"labels": [
{
"type": "subconcept_of",
"value": {
"attributes": {},
"id": "International Law",
"type": "agent",
"value": "International Law",
"labels": []
},
"timestamp": null
}
]
},
"timestamp": null
}

```

### DIFF
--- a/data-in-api/app/repository.py
+++ b/data-in-api/app/repository.py
@@ -28,6 +28,12 @@ from sqlmodel import Session, select
 
 _LOGGER = logging.getLogger(__name__)
 
+LABEL_TYPES_WITH_RELATIONSHIPS = {
+    "case_category",
+    "principal_law",
+    "jurisdiction",
+}  # For now this is a hardcoded set of label types that we known have relationships, i.e Sabin types. Should this list be updated please be aware that different corpora may have the same labels.
+
 
 def check_db_health(db: Session) -> bool:
     """Check database connection health.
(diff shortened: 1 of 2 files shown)


annaCPR (commented):
Question: Does this mean that since the `id` of a label is not guaranteed unique across types, we should be comparing labels based on `id` + `type`?

annaCPR (dismissed):
Lgtm

annaCPR (approved):

PR #1177, opened 2026-04-14
Feature/app 1925 transform litigation filing date

odrakes-cpr:
# What does this do?
Map the family filing date as a label.

## Why is it needed?
We are transforming all relevant values from the Sabin family 

## How was it tested?
Updated transformer tests

### DIFF
--- a/data-in-pipeline/tests/transform/test_navigator_family.py
+++ b/data-in-pipeline/tests/transform/test_navigator_family.py
@@ -1343,7 +1343,7 @@ def test_transform_navigator_family_with_litigation_corpus_type(
                 timestamp=datetime.datetime(2020, 1, 1),
                 value=Label(
                     type="activity_status",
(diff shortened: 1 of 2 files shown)


annaCPR (approved):
Nice

data-in-pipeline/tests/transform/test_navigator_family.py
```diff
@@ -1329,6 +1338,15 @@ def test_transform_navigator_family_with_litigation_corpus_type(
                     type="deprecated_category",
                 ),
             ),
+            LabelRelationship(
+                type="activity_status",
+                timestamp=datetime.datetime(2020, 1, 1),
+                value=Label(
+                    type="activity_status",
+                    id="Filed",
```
jamesgorrie:
```suggestion
                    id="activity_status::Filed",
```

PR #1179, opened 2026-04-14
fix: use type::value for label IDs

jamesgorrie:
# What does this do?
- adds a label ID migration script to move from `{label.value}` => `{label.id}::{label.value}`
- ☝️ does the same in the transformer

I would normally like to split things out, but given the order in which the pipeline runs, this feels like the most pragmatic way to get this in.

- `db_migrations` (updates the existing labels)
- `transform` (sets the IDs to the new format)
- `load` (uses the new format to update the labels in the DB which are now in the new format)

## Why is it needed?

To avoid having conflicts with things like `category::Law` and `entity_type::Law`

## How was it tested?

Pytest

### DIFF
--- a/data-in-pipeline-load-api/app/alembic/migrations/versions/0008_migrate_label_ids_to_type_value.py
+++ b/data-in-pipeline-load-api/app/alembic/migrations/versions/0008_migrate_label_ids_to_type_value.py
@@ -1,0 +1,190 @@
+"""Migrate label IDs from {value} to {type}::{value}
+
+Revision ID: 0008
+Revises: 0007
+Create Date: 2026-04-14
+
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0008"
+down_revision: str | Sequence[str] | None = "0007"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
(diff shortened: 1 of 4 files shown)


odrakes-cpr (approved):

data-in-pipeline-load-api/app/alembic/migrations/versions/0008_migrate_label_ids_to_type_value.py
```diff
@@ -0,0 +1,194 @@
...
+        type_="foreignkey",
+    )
+
+    # Step 2: Update FK columns in link tables to point at the new label IDs.
+    # We JOIN on the old label.id before updating label itself.
+    # The NOT LIKE guard makes this idempotent.
+    op.execute(
+        """
+        UPDATE documentlabelrelationship dlr
+        SET label_id = l.type || '::' || l.id
```
annaCPR, line 50: Question: Can you point to an id that doesn't exist yet? I see that you update the label PKs later.

annaCPR (approved):
Nice one!

data-in-pipeline-load-api/app/alembic/migrations/versions/0008_migrate_label_ids_to_type_value.py
jamesgorrie: @annaCPR could you clarify what you mean? 

> that doesn't exist yet

Which piece of the code are you talking about specifically? `AND l.id NOT LIKE '%::%'`?

PR #1180, opened 2026-04-14
Feature/app 1885 transform organisation metadata to label attributes

annaCPR:
# What does this do?

- Switches from `!=` to `is_distinct_from` when checking label attributes for update.

## Why is it needed?

- If label attributes are NULL in the DB (which happened), the `!=` comparison returns NULL, which in python resolves to False so the attributes are not updated.

## How was it tested?

- Integration tests


odrakes-cpr (approved):
Lgtm

PR #1181, opened 2026-04-16
fix: add International Climate Councils Network to providers

jamesgorrie:
# What does this do?
- add International Climate Councils Network to providers

## Why is it needed?
- we've added this to the `admin-service`


katybaulch (approved):
lgtm

PR #1185, opened 2026-04-21
Fix litigation metadata transform

annaCPR:
# What does this do?

- Transforms missed litigation family metadata: `status`, `id`, `core_object` and `original_case_name` to `attributes`.
- Splits and simplifies the transform test file to improve readability as it had exceeded 3k lines and was getting difficult to navigate

## Why is it needed?

For the migration to new data model - all litigation data needs to be transformed.

## How was it tested?

Unit tests


data-in-pipeline/tests/transform/test_navigator_family_litigation.py
```diff
@@ -0,0 +1,833 @@
...
+            "deprecated_slug": "litigation-family-slug",
+            "identifier::case_number": "CASE-NUMBER 123",
+            "identifier::provider_id": "123456",
+            "published_date": "2020-01-0100:00:00Z",
+            "last_updated_date": "2020-01-0100:00:00Z",
+            "status": "PUBLISHED",
+            "core_object": "Core Object 123",
+            "original_case_name": "Original case name",
+            "case_status": "Decided",
+        },
```
annaCPR, line 472: This is where the new attributes are being tested. Might be easier to see when looking at the individual commits as the rearranging of the test files makes it difficult to review.

odrakes-cpr:
I was actually going to say that we should probably split up the tests now. so nice

odrakes-cpr (approved):

PR #1188, opened 2026-04-21
Feature/app 1809 fix aurora connection issues after credentials rotation

odrakes-cpr:
trying to force a merge to main workflow for infra changes. Stuck in a weird loop where we need permissions to make changes to the cluster. but thats in another pr


annaCPR (approved):
lgtm

PR #1189, opened 2026-04-21
fix: add dates to family_documents

jamesgorrie:
# What does this do?

- adds the date field to `family_documents` in the same way we do for fa `families`


data-in-pipeline/app/transform/navigator_family.py
```diff
@@ -1153,6 +1153,16 @@ def _transform_navigator_document(
     if navigator_document.md5_sum:
         attributes["md5_sum"] = navigator_document.md5_sum
 
+    """
```
katybaulch, line 1156: nitpick: could this be a normal comment?

odrakes-cpr (approved):

annaCPR (approved):
lgtm

katybaulch (approved):
lgtm

data-in-pipeline/app/transform/navigator_family.py
jamesgorrie: [copy/pasta](https://github.com/climatepolicyradar/navigator-backend/blob/e7ce9ca7a34bfe0415b8c8e97e4221454b9a8c3a/data-in-pipeline/app/transform/navigator_family.py#L977-L981), so I'll leave it for consistency if we want to change it later.

I still don't know what the convention is nor how it's enforced.

PR #1194, opened 2026-04-22
fix: remove recently added iam policies related to iam auth on rds

odrakes-cpr:
- we are in a weird muddle state, where the github role permissions required to deploy the recentyly added policies to data in pipeline is out of sync and is erroring when trying to run a pulumi up despite having the requisite roles/permission


annaCPR (approved):
lgtm

PR #1195, opened 2026-04-22
Switch over user credentials on data in load api

odrakes-cpr:
---

### Summary

This PR replaces static database credentials with IAM-based authentication for connecting to our Aurora PostgreSQL instance.

Instead of retrieving a long-lived password from configuration, we now generate a short-lived authentication token using AWS IAM. This improves security posture and aligns with AWS best practices for database access.

---

#### Authentication method

* ❌ Removed usage of:

  * Static username/password from `managed_db_password`
* ✅ Introduced:

  * IAM authentication via `rds.generate_db_auth_token`
  * Username retrieved from SSM Parameter Store
  * Temporary auth token used as the database password

#### AWS integration

* Added AWS session + RDS client usage to generate auth tokens
* Introduced dependency on:

  * SSM Parameter Store (for DB username)
  * IAM permissions for token generation

#### Connection string

* Updated `SQLALCHEMY_DATABASE_URI` to use:

  * IAM auth token instead of static password
  * Same PostgreSQL connection format with SSL enforced

---

### ⚙️ Connection pooling changes

* Updated `pool_recycle`:

  * From **1800s (30 min)** → **840s (14 min)**

This ensures connections are recycled before IAM tokens expire (15 min lifetime), preventing authentication failures from stale connections.

---

### 🧠 Why this matters

* Eliminates long-lived database credentials
* Reduces risk of credential leakage
* Enables fine-grained access control via IAM
* Aligns with AWS-recommended security practices

---

### ⚠️ Operational considerations

* Services must have IAM permissions to call:

  * `rds:GenerateDBAuthToken`
* Database must have IAM authentication enabled
* Connection recycling is critical due to token expiry
* Slight increase in connection overhead due to token generation

### DIFF
--- a/data-in-pipeline-load-api/app/main.py
+++ b/data-in-pipeline-load-api/app/main.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import sys
+from contextlib import asynccontextmanager
 from pathlib import Path
 
 from api import FastAPITelemetry, ServiceManifest, SQLAlchemyTelemetry, TelemetryConfig
@@ -43,7 +44,14 @@ except Exception as _:
 telemetry = FastAPITelemetry(otel_config)
 tracer = telemetry.get_tracer()
 sqlalchemy_telemetry = SQLAlchemyTelemetry(tracer)
-sqlalchemy_telemetry.instrument(get_engine())
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    # Runs once when the server starts — not at import time
+    _LOGGER.info("⚙️ Instrumenting SQLAlchemy telemetry")
+    sqlalchemy_telemetry.instrument(get_engine())
+    yield
 
 
 # Create the FastAPI app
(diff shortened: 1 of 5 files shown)


annaCPR (approved):
lgtm

odrakes-cpr:
Important note

```
Previously, `sqlalchemy_telemetry.instrument(get_engine())` was called
at module level in `main.py`. Because Python executes all module-level
statements at import time, this caused `get_engine()` — and therefore
all downstream AWS calls (SSM parameter lookup, RDS auth token
generation) — to fire the moment any code imported `app`.

In the test environment no AWS credentials are present, so collection
failed with `NoCredentialsError` before a single test could run.

The fix moves the instrumentation call into a FastAPI lifespan handler,
which only executes when the server actually starts. Imports are now
side-effect free, and tests can safely import `app` and override
dependencies without needing AWS credentials.
```

PR #1199, opened 2026-04-22
Add iam authentication data in api

odrakes-cpr:
# What does this do?
 
Move from master user and secrets to using the iam authentication. 

We have created the relevant policies in the data in pipeline infra. So here all we need to do is switch out the db user for the reader db. (Note this user has already been created in aurora). 

It follows the same process for the data-in-pipeline-load-api.

As we pull through the majority of infra code from the data-in-pipeline not much is required as IOC, just the pulumi secrets for the user. 

## Why is it needed?

This is a more robust way of accessing our db, plus this prevents us having to re dploy every week when our credentials are rotated. 

## How was it tested?
n/a


data-in-api/infra/__main__.py
```diff
@@ -231,13 +233,8 @@
...
                 runtime_environment_secrets={
                     "LOAD_DATABASE_URL": aurora_read_replica_db_url_parameter.arn,
-                    "MANAGED_DB_PASSWORD": aurora_read_replica_db_secret_arn,
-                    "DB_MASTER_USERNAME": aurora_read_replica_db_username_parameter.arn,
-                    # TODO: ^ to be removed in a later PR in favour of 👇
                     "DB_URL": aurora_read_replica_db_url_parameter.arn,
                     "DB_NAME": aurora_read_replica_db_name_parameter.arn,
-                    "DB_USERNAME": aurora_read_replica_db_username_parameter.arn,
-                    # This is in the format `{"password": "xxx", "username": "xxx"}`
                     "DB_SECRETS": aurora_read_replica_db_secret_arn,
```
annaCPR, line 238: I think we can remove this as well as it was only used to connect here

data-in-api/app/settings.py
annaCPR, line 29: This can also be removed as it should no longer be used anywhere.

annaCPR (approved):
A couple of things to tidy but non-blocking. Great job on this!

PR #1202, opened 2026-04-23
fix: last bits of iam auth for aurora

jamesgorrie:
Hooks up the last bits for `IAM` auth to `data-in-pipeline` Aurora.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -129,6 +129,7 @@ jobs:
     steps:
       - name: Fail if PR touches both app code and infra
         run: |
+          exit 0
           SERVICES='${{ needs.changes.outputs.services }}'
           INFRA='${{ needs.changes.outputs.infra_services }}'
           # Empty or '[]' means no changes; only fail when BOTH have actual content
--- a/data-in-api/app/session.py
+++ b/data-in-api/app/session.py
@@ -8,6 +8,7 @@ sessions. Use get_db_context() for all database operations.
 import logging
 from collections.abc import Generator
 from contextlib import contextmanager
+from urllib.parse import quote_plus
 
 from sqlalchemy.engine import Engine
 from sqlmodel import Session, create_engine
@@ -60,7 +61,7 @@ def get_db() -> Generator[Session, None, None]:
 def _build_database_uri() -> str:
     """Fetch the IAM auth token and build the SQLAlchemy URI."""
 
-    username = get_ssm_parameter("data-in-pipeline-aurora-read-replica-db-username")
+    username = get_ssm_parameter("/data-in-pipeline/aurora/read-replica-db-username")
 
     aws_session = get_aws_session()
     rds_client = aws_session.client("rds")
@@ -76,7 +77,7 @@ def _build_database_uri() -> str:
     )
 
     return (
-        f"postgresql://{username}:{token}@"
+        f"postgresql://{username}:{quote_plus(token)}@"
         f"{endpoint}:{port}/{settings.db_name.get_secret_value()}"
         f"?sslmode={settings.db_sslmode}"
     )
--- a/data-in-api/app/settings.py
+++ b/data-in-api/app/settings.py
@@ -41,7 +41,7 @@ class Settings(BaseSettings):
     # 'prefer' tries SSL but falls back to non-SSL for local dev (validates
     # certs if SSL is used). For production RDS without cert validation,
     # set db_sslmode=require via environment variable.
-    db_sslmode: str = "prefer"
+    db_sslmode: str = "require"
 
 
 # Pydantic settings are set from the env variables passed in via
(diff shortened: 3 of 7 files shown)


.github/workflows/pull-request.yml
```diff
@@ -129,6 +129,7 @@ jobs:
     steps:
       - name: Fail if PR touches both app code and infra
         run: |
+          exit 0
```
jamesgorrie, line 132: I'll put this back once deployed

data-in-api/app/session.py
```diff
@@ -60,7 +61,7 @@ def get_db() -> Generator[Session, None, None]:
 def _build_database_uri() -> str:
     """Fetch the IAM auth token and build the SQLAlchemy URI."""
 
-    username = get_ssm_parameter("data-in-pipeline-aurora-read-replica-db-username")
+    username = get_ssm_parameter("/data-in-pipeline/aurora/read-replica-db-username")
```
jamesgorrie, line 64: [This is the name](https://github.com/climatepolicyradar/navigator-backend/blob/df7b7e024d05b273829c76d17b2eb0cd9d6f7339/data-in-pipeline/infra/__main__.py#L540-L544) - you could export the config name if you wanted and read it here - but I think the namespacing is clear enough

data-in-api/app/session.py
```diff
@@ -76,7 +77,7 @@ def _build_database_uri() -> str:
     )
 
     return (
-        f"postgresql://{username}:{token}@"
+        f"postgresql://{username}:{quote_plus(token)}@"
```
jamesgorrie, line 80: I am not 100% sure this is needed as it was part of the whack a mole, but in the name of "It works!" I am keeping it.

data-in-api/app/settings.py
```diff
@@ -41,7 +41,7 @@ class Settings(BaseSettings):
     # 'prefer' tries SSL but falls back to non-SSL for local dev (validates
     # certs if SSL is used). For production RDS without cert validation,
     # set db_sslmode=require via environment variable.
-    db_sslmode: str = "prefer"
+    db_sslmode: str = "require"
```
jamesgorrie, line 44: This is required by IAM and is best practice, you can set it as an env var locally, either in .env or the docker-compose file if we need something different.

data-in-api/infra/__main__.py
```diff
@@ -120,6 +120,10 @@
     "aurora-read-replica-security-group-id"
 )
 
+aurora_cluster_resource_id = data_in_pipeline_stack.require_output(
+    "aurora-cluster-resource-id"
+)
+
```
jamesgorrie, line 126: This is a duplicate of another value that is namespaced with the stack and other values, so I did this for ease of use (and is why I don't namespace things with environments)

data-in-api/infra/__main__.py
```diff
@@ -154,6 +160,13 @@
                         ],
                         resources=[args["aurora_read_replica_db_secret_arn"]],
                     ),
+                    aws.iam.GetPolicyDocumentStatementArgs(
+                        effect="Allow",
+                        actions=["rds-db:connect"],
+                        resources=[
+                            f"arn:aws:rds-db:eu-west-1:{account_id}:dbuser:{args['aurora_cluster_resource_id']}/{args['db_username']}"
```
jamesgorrie, line 167: Thanks Claude

data-in-pipeline/infra/__main__.py
```diff
@@ -738,8 +739,11 @@ def generate_secret_key(project: str, aws_service: str, name: str):
                     "DB_NAME": data_in_pipeline_aurora_write_replica_db_name.arn,
                     # This is in the format `{"password": "xxx", "username": "xxx"}`
                     "DB_SECRETS": data_in_pipeline_aurora_write_replica_db_secrets.secret_arn,
+                    "DB_USERNAME": data_in_pipeline_aurora_write_replica_db_username.arn,
+                    "MANAGED_DB_PASSWORD": data_in_pipeline_load_api_cluster_password_secret.secret_arn,
```
jamesgorrie, line 743: I needed to put these back as the load-api is not spinning up - please adjust as we deem fit.

data-in-pipeline/infra/grants.sql
```diff
@@ -0,0 +1,4 @@
+GRANT USAGE ON SCHEMA public TO api_data_reader;
+GRANT SELECT ON ALL TABLES IN SCHEMA public TO api_data_reader;
+ALTER DEFAULT PRIVILEGES IN SCHEMA public
+  GRANT SELECT ON TABLES TO api_data_reader;
```
jamesgorrie, line 4: I had to run this on the box via the bastion as I couldn't select any table. @odrakes-cpr can we find a way to not do this or make this programatic when a box spins up?

jamesgorrie: @odrakes-cpr we also might need to do this in code.

jamesgorrie:
@odrakes-cpr I am merging this in to get systems up again.

I don't mean to push through someones work, but it was blocking changes upstream that we needed to get working as we were user testing.

PR #1209, opened 2026-04-23
App 1809 fix aurora iam for data-in-api

annaCPR:
# What does this do?

- Implements IAM authentication (again) for the `data-in-api` with proper engine and IAM auth token management which creates the engine at module level but only generates and injects the token into the connection string on connection to DB. This way the tokens are properly refreshed before expiry.

## Why is it needed?

So we can connect to Aurora with a dedicated read-only user and via IAM authentication rather than using master credentials which are rotated weekly and were causing service errors.

## How was it tested?

The exact same implementation has already rolled out and thoroughly tested on the `load-api`.

DRIVE-BY:

- removes an unused pytest fixture (looks like it was left over when db migration code was moved to the load api) 
- sets `UV_PROJECT_ENVIRONMENT` variable in the Dockerfile for both `load-api` and `data-in-api` (it follows a pattern from DIP and resolves the issue whereby running tests for either service in docker locally breaks the venv)
- updates the test DB engine fixture to use username and password from env variables provided in docker-compose rather than Settings.py so that we can bypass the IAM auth token generation and run tests
- renames some of the variables in Session.py for the `load-api` to match the `data-in-api`

### DIFF
--- a/data-in-api/app/session.py
+++ b/data-in-api/app/session.py
@@ -41,7 +41,7 @@ _engine = create_engine(
     pool_pre_ping=True,  # Verify connections before use
     pool_size=10,  # Base connection pool size
     max_overflow=100,  # Additional connections when pool exhausted
-    pool_recycle=840,  # Recycle every 14 min to avoid expired IAM auth tokens (15 min lifetime). https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html
+    pool_recycle=600,  # Recycle every 10 min to avoid expired IAM auth tokens (15 min lifetime). https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html
     pool_timeout=30,  # Wait up to 30s for a connection before error
     isolation_level="READ COMMITTED",  # PostgreSQL default, explicit
     connect_args={"options": f"-c statement_timeout={settings.statement_timeout}"},
--- a/data-in-api/app/session.py
+++ b/data-in-api/app/session.py
@@ -8,9 +8,11 @@ sessions. Use get_db_context() for all database operations.
 import logging
 from collections.abc import Generator
 from contextlib import contextmanager
+from typing import Any
 
 from sqlalchemy import event
 from sqlalchemy.engine import Engine
+from sqlalchemy.engine.interfaces import Dialect
 from sqlmodel import Session, create_engine
 
 from app.aws import get_aws_session
@@ -89,6 +91,18 @@ def get_db() -> Generator[Session, None, None]:
 
 
 def _generate_token() -> str:
+    """Generates a short-lived IAM authentication token for Aurora.
+
+    Calls the AWS RDS API to obtain a signed token that can be used as the
+    password when connecting to Aurora. The token is valid for 15
+    minutes and is scoped to the configured user, host, port, and region.
+
+    Intended for use inside the ``do_connect`` event listener so each new
+    DB connection authenticates with a fresh, non-expired token.
+
+    :return: A signed IAM auth token suitable for use as a Postgres password.
+    :rtype: str
+    """
     aws_session = get_aws_session()
     rds_client = aws_session.client("rds")
     return rds_client.generate_db_auth_token(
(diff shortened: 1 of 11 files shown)


data-in-api/app/session.py
```diff
@@ -9,34 +9,29 @@
...
-class DBSecrets(BaseModel):
-    username: str
-    password: str
-
-
-db_secrets = DBSecrets.model_validate_json(settings.db_secrets.get_secret_value())
+endpoint = settings.db_url.get_secret_value()
+port = settings.db_port
+region = settings.aws_region
+username = settings.db_username.get_secret_value()
```
jamesgorrie: nit: keeping these things named the same thing across their lifecycle helps with reducing cognitive load when parsing them

data-in-api/app/session.py
```diff
@@ -46,7 +41,7 @@ class DBSecrets(BaseModel):
     pool_pre_ping=True,  # Verify connections before use
     pool_size=10,  # Base connection pool size
     max_overflow=100,  # Additional connections when pool exhausted
-    pool_recycle=1800,  # Recycle connections after 30 minutes
+    pool_recycle=840,  # Recycle every 14 min to avoid expired IAM auth tokens (15 min lifetime). https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html
```
jamesgorrie: suggestion: I'd probably go for something a little more different to avoid latency issues and the like as this is prime for a race condition. 10 minutes?

data-in-api/app/session.py
```diff
@@ -93,12 +88,25 @@ def get_db() -> Generator[Session, None, None]:
...
+    rds_client = aws_session.client("rds")
+    return rds_client.generate_db_auth_token(
+        DBHostname=endpoint,
+        Port=port,
+        DBUsername=username,
+        Region=region,
+    )
+
+
+@event.listens_for(_engine, "do_connect")
```
jamesgorrie, line 116: for my knowledge - what is this plugging into?

jamesgorrie (commented):
nice! Have you seen this working in a AWS environment?

data-in-api/app/session.py
annaCPR: So this is adding an event listener to the _engine which it accepts as a parameter. When a `do_connect` event is emitted, it triggers this function which then generates the IAM token. This is the SQAlchemy recommended approach: https://docs.sqlalchemy.org/en/20/core/engines.html#generating-dynamic-authentication-tokens

annaCPR: Added a proper docstring with the link.

annaCPR:
@jamesgorrie as per PR description - this implementation is already deployed and working correctly for the `load-api`.

data-in-api/app/session.py
jamesgorrie: Thanks! Super interesting pattern.
Could be worth dropping that document link in.

data-in-api/app/session.py
```diff
@@ -93,12 +90,56 @@ def get_db() -> Generator[Session, None, None]:
...
+    aws_session = get_aws_session()
+    rds_client = aws_session.client("rds")
+    return rds_client.generate_db_auth_token(
+        DBHostname=db_url,
+        Port=db_port,
+        DBUsername=db_username,
+        Region=aws_region,
+    )
+
+
```
jamesgorrie, line 115:
```suggestion

# @see:  https://docs.sqlalchemy.org/en/20/core/engines.html#generating-dynamic-authentication-tokens
```

data-in-api/tests/conftest.py
```diff
@@ -37,17 +34,12 @@ def engine():
...
-    db_host = settings.db_url.get_secret_value()
-    db_port = settings.db_port
-    db_name = settings.db_name.get_secret_value()
-    db_username = db_secrets.username
-    db_password = db_secrets.password
-
-    # Parse password from JSON if needed
-    if db_password is not None and db_password.startswith("{"):
-        db_password = json.loads(db_password)["password"]
+    db_host = os.getenv("DB_URL")
```
jamesgorrie: note: having these names change through their lifetime can often create cognitive overload for future travellers to the code and make refactoring a little harder.

data-in-api/tests/conftest.py
```diff
@@ -37,17 +34,12 @@ def engine():
...
+    db_port = os.getenv("DB_PORT")
+    db_name = os.getenv("DB_NAME")
+    db_username = os.getenv("DB_USERNAME")
+    db_password = os.getenv("DB_MASTER_PASSWORD")
```
jamesgorrie, line 42: question: why do we still need the password?

jamesgorrie (approved):
very very nice

data-in-api/tests/conftest.py
annaCPR: Because that is currently the easiest way to connect to the test DB while circumventing IAM. I have an idea of how we can get around it (which will also allow us to run the service locally) but this will come in a separate PR

data-in-api/tests/conftest.py
annaCPR: Missed that - will correct

data-in-api/app/session.py
annaCPR: It's already in the docstring so I don't see why we need a separate comment

PR #1210, opened 2026-04-27
Schedule DIP to run daily at 5am

annaCPR:
# What does this do?

- Adds a schedule to the DIP deployment so that the flow runs every day at 5am (after the document pipeline finishes). 

## Why is it needed?

- So that the `data-in-api` has its data automatically updated


katybaulch (approved):
lgtm

PR #1211, opened 2026-04-28
App 1809 fix aurora iam

annaCPR:
# What does this do?

- Adds `DB_SSLMODE` env var for `data-in-api` and `load-api` and sets both to "require" (as per comment in the Settings.py)
- Fixes db conection string for `data-in-api`

## Why is it needed?

- Connection to the Aurora DB was failing

### DIFF
--- a/data-in-api/app/session.py
+++ b/data-in-api/app/session.py
@@ -33,7 +33,7 @@ SQLALCHEMY_DATABASE_URI = (
 _LOGGER.info(
     f"🔌 Initialising database engine for "
     f"{settings.db_url.get_secret_value()}:"
-    f"{settings.db_port}/{settings.db_name}"
+    f"{settings.db_port}/{settings.db_name.get_secret_value()}"
 )
 
 # Engine with connection pooling to prevent connection leaks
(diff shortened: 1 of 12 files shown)


data-in-api/app/session.py
```diff
@@ -26,7 +26,7 @@
 db_username = settings.db_username.get_secret_value()
 
 SQLALCHEMY_DATABASE_URI = (
-    f"postgresql://{db_username}@{db_url}:{db_port}/{settings.db_name}"
+    f"postgresql://{db_username}@{db_url}:{db_port}/{settings.db_name.get_secret_value()}"
```
jamesgorrie, line 29: Interesting - I thought accessing secrets as plaintext would have broken startup.

jamesgorrie (approved):

data-in-api/app/session.py
annaCPR: We only connect to the DB on request so the application can start correctly

annaCPR: Annoyingly, the tests would have picked that up previously because they were actually using Settings.py but I had to remove that temporarily as I didn't have a workaround for fetching an auth token so defaulted back to using username and password. I'm hoping to put that back in soon

PR #1212, opened 2026-04-28
App 1809 fix aurora iam

annaCPR:
# What does this do?

- Removes all environment variables related to the master Aurora credentials and any policies that give permissions to them from `data-in-api` and `load-api` infra

## Why is it needed?

- We are no longer using master credentials to connect to Aurora (we have enabled IAM auth and created unique users for each service).

### DIFF
--- a/data-in-api/infra/__main__.py
+++ b/data-in-api/infra/__main__.py
@@ -112,10 +112,6 @@ aurora_read_replica_db_username_parameter = data_in_pipeline_stack.get_output(
     "aurora-read-replica-db-username-parameter-name"
 ).apply(lambda name: aws.ssm.get_parameter(name=name))
 
-aurora_read_replica_db_secret_arn = data_in_pipeline_stack.get_output(
-    "aurora-read-replica-db-secrets-secret-arn"
-)
-
 aurora_read_replica_db_security_group_id = data_in_pipeline_stack.get_output(
     "aurora-read-replica-security-group-id"
 )
@@ -132,7 +128,6 @@ instance_role_policy = aws.iam.RolePolicy(
         aurora_read_replica_db_url_parameter=aurora_read_replica_db_url_parameter.arn,
         aurora_read_replica_db_name_parameter=aurora_read_replica_db_name_parameter.arn,
         aurora_read_replica_db_username_parameter=aurora_read_replica_db_username_parameter.arn,
-        aurora_read_replica_db_secret_arn=aurora_read_replica_db_secret_arn,
         aurora_cluster_resource_id=aurora_cluster_resource_id,
         db_username=aurora_read_replica_db_username_parameter.value,
     ).apply(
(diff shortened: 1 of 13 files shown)


data-in-api/infra/__main__.py
```diff
@@ -245,15 +232,9 @@
         image_repository=aws.apprunner.ServiceSourceConfigurationImageRepositoryArgs(
             image_configuration=aws.apprunner.ServiceSourceConfigurationImageRepositoryImageConfigurationArgs(
                 runtime_environment_secrets={
-                    "LOAD_DATABASE_URL": aurora_read_replica_db_url_parameter.arn,
-                    "MANAGED_DB_PASSWORD": aurora_read_replica_db_secret_arn,
-                    "DB_MASTER_USERNAME": aurora_read_replica_db_username_parameter.arn,
-                    # TODO: ^ to be removed in a later PR in favour of 👇
```
jamesgorrie, line 251: Note and validation: the other arns are covered by [these `actions`](https://github.com/climatepolicyradar/navigator-backend/pull/1212/changes#diff-27e3ea3e2d4ed0b716fef9efe644eddce3bdb6e49b80db19df8b59f32ba9e5ccL145-L147)?

```
"ssm:GetParameter",
"ssm:GetParameters",
"ssm:DescribeParameters",
```

data-in-api/infra/__main__.py
```diff
@@ -245,15 +232,9 @@
         image_repository=aws.apprunner.ServiceSourceConfigurationImageRepositoryArgs(
             image_configuration=aws.apprunner.ServiceSourceConfigurationImageRepositoryImageConfigurationArgs(
                 runtime_environment_secrets={
-                    "LOAD_DATABASE_URL": aurora_read_replica_db_url_parameter.arn,
-                    "MANAGED_DB_PASSWORD": aurora_read_replica_db_secret_arn,
-                    "DB_MASTER_USERNAME": aurora_read_replica_db_username_parameter.arn,
```
jamesgorrie, line 250: question: Have these been removed from all the Pydantic settings in the apps? If not they won't start up as those fail at startup.

[It looks like it - albeit the tests are using them](https://github.com/climatepolicyradar/navigator-backend/blob/a7891d811682c2709533a38e5e690d651a1f0a83/data-in-pipeline-load-api/tests/conftest.py#L35) - so should we remove those as part of this?

annaCPR: They have been already removed from Settings.py. If I remove them from the test fixtures, I will have to also fix all the tests to mock the IAM token generation and I'd rather do that in a separate PR.

data-in-api/infra/__main__.py
annaCPR: Can you elaborate please James?

jamesgorrie: we're removing `secretsmanager:GetSecretValue` and `secretsmanager:DescribeSecret` - just double checking this doesn't mess with these values.

I don't think they do, but you asked for a fine tooth comb.

jamesgorrie (approved):
Annoyingly there seems to be a bug in the no-infra-and-code check.

data-in-api/infra/__main__.py
annaCPR: Yes I did, thank you 🙏  These permissions were scoped to a single resource that we're removing anyway: `resources=[args["aurora_read_replica_db_secret_arn"]]`

PR #1213, opened 2026-04-28
Do not return Failure if parent concept is missing

annaCPR:
# What does this do?

- Removes code that returned a Failure result if a litigation concept on a family had a reference to a non-existent parent concept.

## Why is it needed?

- This meant that the whole document was not getting transformed which has led to stale data.


jamesgorrie (approved):

jamesgorrie:
@annaCPR we could leave the log to investigate / report on which aren't coming through while not failing the run?

annaCPR:
I wanted to but the Failure bubbles up all the way to the top and I didn't have an idea of how to do this nicely without blocking Jess. I've left the reporting code for now so we can easily add this back in after we update the data.

PR #1214, opened 2026-04-28
fix: lowercase `published`

jamesgorrie:
# What does this do?

- lowercases published universally

This is currently already the case in 
- [`families-api`](https://api.climatepolicyradar.org/families/CCLW.family.i00005745.n0000) (see `documents[].document_status`)
- [`data-in-api`](https://api.climatepolicyradar.org/data-in/documents/CCLW.legislative.4355.1972) (see `attributes.status`)

## Why is it needed?

- we currently aren't matching published on families

## How was it tested?

pytest

### DIFF
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -1000,10 +1000,10 @@ def _transform_navigator_family(
     """
 
     contains_published_document = [
-        doc for doc in navigator_family.documents if doc.document_status == "PUBLISHED"
+        doc for doc in navigator_family.documents if doc.document_status == "published"
     ]
     if navigator_family.documents and contains_published_document:
-        attributes["status"] = "PUBLISHED"
+        attributes["status"] = "published"
 
     return Success(
         Document(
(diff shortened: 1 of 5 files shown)


data-in-pipeline/app/transform/navigator_family.py
```diff
@@ -1000,10 +1000,10 @@ def _transform_navigator_family(
     """
 
     contains_published_document = [
-        doc for doc in navigator_family.documents if doc.document_status == "PUBLISHED"
+        doc for doc in navigator_family.documents if doc.document_status == "published"
```
annaCPR, line 999: As far as I know the values in RDS are uppercase so this should break?

jamesgorrie: @annaCPR These values are read from the [`families-api`](https://api.climatepolicyradar.org/families/CCLW.family.i00005745.n0000) <- which are lowercase.

jamesgorrie: @annaCPR [This seems to be the code that lowercases it](https://github.com/climatepolicyradar/navigator-backend/blob/main/families-api/app/models.py#L441)

annaCPR: I keep forgetting we're reading from `/families` and not RDS directly 🙄

annaCPR (approved):
lgtm

PR #1217, opened 2026-04-30
Fix memory leak when fetching all families

annaCPR:
# What does this do?

- Moves the list of `validated_families` to inside the `while` loop to stop it growing exponentially
- Changes the type of `raw_payload` to one matching what is being returned by the Http client

## Why is it needed?

- We were seeing memory leaks in the Prefect flow when trying to save extract result to S3 because the list of results was extremely large.
- `raw_payload` was being passed an already serialised pydantic object rather than raw data from the Http response

## How was it tested?
Unit tests

### DIFF
--- a/data-in-pipeline/tests/test_family_pipeline.py
+++ b/data-in-pipeline/tests/test_family_pipeline.py
@@ -102,7 +102,7 @@ def test_process_family_updates_flow_multiple_families(  # noqa: PLR0913
 
     envelope_1 = ExtractedEnvelope(
         data=page_1_data,
-        raw_payload=page_1_data,
+        raw_payload=[family.model_dump() for family in page_1_data],
         id="test-uuid-1",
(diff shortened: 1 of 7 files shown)


data-in-pipeline/tests/test_family_pipeline.py
```diff
@@ -338,7 +338,7 @@ def test_etl_pipeline_partial_transformation_failure(  # noqa: PLR0913
 
     envelope = ExtractedEnvelope(
         data=[valid_family, invalid_family],
-        raw_payload=[valid_family, invalid_family],
+        raw_payload=[family.model_dump() for family in [valid_family, invalid_family]],
```
katybaulch, line 341: praise: yay for list comprehensions!

katybaulch (approved):
nice spots, well done!

PR #1218, opened 2026-04-30
Allow families that fail event metadata validations to be transformed

annaCPR:
# What does this do?

- Changes the type of `NavigatorEvent.metadata` from `dict[str, list[str]` => `dict[str, list[str] | str]`

## Why is it needed?

- We have **243** families that fail model validation due to `NavigatorEvent.metadata.even_type` being a string rather than a list. 
Screenshot 2026-04-30 at 16 06 27

From RDS:
Screenshot 2026-04-30 at 16 07 39

We don't use the value of `NavigatorEvent.metadata.even_type` as it is a duplicate of `NavigatorEvent.event_type` so we are allowing str values here so that the affected families can still be transformed.

### DIFF
--- a/data-in-pipeline/app/extract/connectors.py
+++ b/data-in-pipeline/app/extract/connectors.py
@@ -2,7 +2,7 @@ import datetime
 from http import HTTPStatus
 
 import requests
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 from requests.adapters import HTTPAdapter, Retry
 from returns.result import Failure, Result, Success
 
@@ -82,8 +82,8 @@ class NavigatorFamily(BaseModel):
     slug: str
     concepts: list[NavigatorConcept] = []
     metadata: dict[str, list[str]] = {}
-    published_date: str | None
-    last_updated_date: str | None
+    published_date: str | None = None
+    last_updated_date: str | None = None
 
 
 class PageFetchFailure(BaseModel):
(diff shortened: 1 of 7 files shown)


data-in-pipeline/app/extract/connectors.py
```diff
@@ -17,7 +17,12 @@ class NavigatorEvent(BaseModel):
     import_id: str
     event_type: str
     date: datetime.datetime
-    metadata: dict[str, list[str]] = {}
+    # We have quite a few families that fail model validation due to
+    # event.metadata.even_type being a string rather than a list.
+    # We don't use the value of event.metadata.even_type as it is a duplicate
+    # of event.event_type so we are allowing str values here so that the
+    # affected families can still be transformed.
+    metadata: dict[str, list[str] | str] = {}
```
odrakes-cpr, line 26: I am a bit confused on the type here because it seems like it's a list of strings coming out of the families api.

```
"events": [
{
"import_id": "Sabin.event.98781.n0000",
"title": "Filing Year For Action",
"date": "2024-01-01T00:00:00Z",
"event_type": "Filing Year For Action",
"status": "OK",
"metadata": {
"event_type": [
"Filing Year For Action"
],
"description": [
"Filing Year For Action"
],
"action_taken": [],
"datetime_event_name": [
"Filing Year For Action"
]
}
}
],
```

odrakes-cpr:
```
{
"import_id": "CCLW.event.i00009110.n0000",
"title": "Approved",
"date": "2023-01-01T00:00:00Z",
"event_type": "Passed/Approved",
"status": "OK",
"metadata": {
"event_type": [
"Passed/Approved"
],
"datetime_event_name": [
"Passed/Approved"
]
}
}
```

annaCPR: It returns correct data for most families but not all:
https://api.climatepolicyradar.org/families/CCLW.family.i00002209.n0000

odrakes-cpr: ignore me - you've sent an example where this happens. https://api.climatepolicyradar.org/families/CCLW.family.i00002209.n0000

odrakes-cpr (approved):

PR #1220, opened 2026-05-05
Small QoL improvements for CSV download endpoint

katybaulch:
# What does this do?

- Remove exploding info/debug logs in loop
- Avoid repeated expensive set/list building in inner loops
- Stop materialising giant CSV strings in memory first only

## Why is it needed?

Hopefully reduce latency & CPU utilisation (even if only briefly)

## How was it tested?

New tests

### DIFF
--- a/backend-api/app/service/download.py
+++ b/backend-api/app/service/download.py
@@ -5,7 +5,7 @@ import zipfile
 from collections import defaultdict
 from io import BytesIO, StringIO
 from logging import getLogger
-from typing import Any, Mapping, Optional, Sequence, cast
+from typing import Any, Iterator, Mapping, Optional, Sequence, cast
 
 import pandas as pd
 from db_client.models.dfce import (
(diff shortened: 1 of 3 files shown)


backend-api/app/service/download.py
```diff
@@ -151,6 +151,32 @@ def parse_concept_labels(concept_labels: Sequence[str], prefix: str) -> str:
     )
 
 
+def _get_matching_document_import_ids(
+    search_response_families: Sequence[SearchResponseFamily],
+    documents_by_family_slug: Mapping[str, Sequence[FamilyDocument]],
+) -> set[str]:
+    """Return document import IDs that match search passage hits."""
```
annaCPR: Missing params and return in docstring

backend-api/app/service/download.py
```diff
@@ -330,24 +371,36 @@ def process_result_into_csv(
...
+@observe("process_result_into_csv")
+def process_result_into_csv(
+    db: Session,
+    search_response_families: Sequence[SearchResponseFamily],
+    base_url: str | None,
+    is_browse: bool,
+    theme: str | None = None,
+) -> str:
+    """Process a search/browse result into a CSV string."""
+    return b"".join(
```
annaCPR, line 416: Also missing params and return in docstring

annaCPR (approved):
lgtm

backend-api/app/service/download.py
katybaulch: Fixed

PR #1221, opened 2026-05-05
Fix return transformation errors

odrakes-cpr:
## Summary

Refactor the family transformation pipeline to use the **Validation pattern**: `Result` is preserved at the public boundary for fatal failures, and non-fatal issues are accumulated as typed `TransformWarning`s carried inside the success value. Transforms still produce their output when sub-steps have to skip something — callers observe what happened via the warnings list rather than the whole family being dropped.

This re-rolls the earlier "drop `Result` entirely" attempt after PR feedback that it gave up too much of the `returns` library's ergonomics.

## What changed

- **New `TransformOutput`** Pydantic model with `documents`, `collection_documents`, and `warnings`. Replaces the unnamed `tuple[list[Document], list[Document]]` shape and gives warnings a place to live.
- **New discriminated-union `TransformWarning`** = `UnknownParentLabel | UnknownGeography`, keyed by a `kind` literal. Defined in `app.transform.models` alongside `CouldNotTransform` so the warning vocabulary has a single source of truth.
- **`transform()` and `transform_navigator_family()` still return `Result[TransformOutput, CouldNotTransform]`.** Callers keep `.bind`, `.map`, structural pattern matching, and a stable return shape.
- **Internal helpers drop `Result` where they have no fatal path.** `_transform_navigator_family`, `_transform_navigator_document`, `_transform_litigation_concepts_to_label_relationships`, `_transform_geographies`, and `_transform_litigation_data` now return `tuple[..., list[TransformWarning]]`. This is honest about what they actually do — accumulate skips, don't fail.
- **Litigation concepts:** missing parent labels are captured as `UnknownParentLabel` warnings instead of being silently dropped.
- **Geographies:** the previous `else` branch logged `"Geography not found"` for the `XAA` sentinel, which is the *intentional* exclusion. Now `XAA` is silently skipped (correct) and genuine unknown IDs become `UnknownGeography` warnings.
- **Wrapper cleanup.** The redundant `match Success() / Failure()` rewrap in `transform_navigator_family` is gone. The unreachable `Failure(NoMatchingTransformations())` branch and its import are removed.
- **Pipeline-level caller** (`@task transform`) widens its return to `tuple[list[Document], list[Exception | TransformWarning]]`. Warnings flow into the same bucket as errors so existing downstream handling keeps working; per-warning structured logging is added on the success branch for observability.

## Why

- **Honest types without losing `Result`.** `Result` was used everywhere — including in helpers with no genuine failure path — so readers had to scan dead `Failure` branches to find the real logic. The new shape uses `Result` only where fatal failure actually exists.
- **Observability of soft failures.** Missing parent concepts and unknown geography IDs were silently dropped or logged in ways that lost context. They're now first-class, typed data threaded through the return.
- **Partial success.** A single bad litigation concept or one unknown geography no longer needs special-case handling — the document still flows through and the issue is reported alongside it.
- **Stable boundary for callers.** Public functions return `Result` whether or not there's currently a fatal path, so adding future fatal cases is additive rather than signature-changing.

### DIFF
--- a/data-in-pipeline/app/transform/models.py
+++ b/data-in-pipeline/app/transform/models.py
@@ -1,3 +1,8 @@
+from typing import Annotated, Literal
+
+from pydantic import BaseModel, Field
+
+
 class CouldNotTransform(Exception):
     """Raised when a transformer function cannot be found for the given input."""
 
@@ -8,3 +13,47 @@ class NoMatchingTransformations(Exception):
     """Raised when the transformer finds no matching transformations for the given input."""
 
     pass
+
+
+# ---------------------------------------------------------------------------
+# Warnings (non-fatal)
+# ---------------------------------------------------------------------------
+# A warning means "I produced a Document, but here's something I had to skip
+# or couldn't resolve." The transform still completes; warnings are an audit
+# log so callers can observe and report what happened. Compare with
+# `CouldNotTransform`, which is used at the boundary for genuinely fatal
+# failures where no Document can be produced at all.
+
+
+class UnknownParentLabel(BaseModel):
+    """A litigation concept referenced a parent label that wasn't in the input.
+
+    The child label is still emitted; only the unresolved `subconcept_of`
+    link is dropped.
+    """
+
+    kind: Literal["unknown_parent_label"] = "unknown_parent_label"
+    family_import_id: str
+    relation: str
+    parent_name: str
+
+
+class UnknownGeography(BaseModel):
+    """A geography ID was supplied that isn't in the geographies lookup table.
+
+    The geography is silently skipped; this warning surfaces it for inspection.
+    Note: the sentinel `XAA` ("No Geography") is intentionally excluded and
+    does NOT produce a warning.
+    """
+
+    kind: Literal["unknown_geography"] = "unknown_geography"
+    family_import_id: str
+    geography_id: str
+
+
+# Discriminated union — `kind` is the discriminator field.
+# @see: https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions
+type TransformWarning = Annotated[
+    UnknownParentLabel | UnknownGeography,
+    Field(discriminator="kind"),
+]
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -84,10 +84,6 @@ def transform_navigator_family(
             f"Transform completed: produced {len(result.documents)} documents, "
             f"{len(result.errors)} errors"
         )
-        if result.errors:
-            logger.warning(
-                f"Transformation produced {len(result.errors)} non-fatal errors"
-            )
         return result
 
 
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -7,7 +7,7 @@ from data_in_models.models import (
     LabelRelationship,
     LabelWithoutDocumentRelationships,
 )
-from returns.result import Failure, Result, Success
+from pydantic import BaseModel, ConfigDict
 
 from app.bootstrap_telemetry import get_logger, log_context
 from app.extract.connectors import (
@@ -17,7 +17,7 @@ from app.extract.connectors import (
 )
 from app.geographies import geographies_lookup
 from app.models import Identified, NavigatorConcept
-from app.transform.models import CouldNotTransform, NoMatchingTransformations
+from app.transform.models import CouldNotTransform
 
 mcf_projects_corpus_import_ids = [
     "MCF.corpus.AF.n0000",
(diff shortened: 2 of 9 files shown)


data-in-pipeline/app/transform/navigator_family.py
```diff
@@ -59,23 +59,32 @@
...
 
+class TransformResult(BaseModel):
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True
+    )  # allowing non-Pydantic types in this model
+    documents: list[Document]
+    errors: list[CouldNotTransform]
+
+
+class LitigationConceptLabelTransformResult(BaseModel):
```
annaCPR: Question: Do we need specific types for label errors and document errors? Can we not make something more generic to capture either?

jamesgorrie (commented):
An overarching thought here

Is using `tuple[Success, Error]` the same as using the `Result` type, but with a lot of very useful functionality e.g. better type matching, and more descriptive typing missing?

I think I might just be lacking clarity on the benefit of shifting from a well-known implementation of `Result` => home-rolled `tuple[Success, Failure]`

Q2: Why would you have a list of errors? Can a transformation fail in multiple ways? All good id so, then having `[Success, list[Failure]]` makes sense

odrakes-cpr:
> An overarching thought here
> 
> Is using `tuple[Success, Error]` the same as using the `Result` type, but with a lot of very useful functionality e.g. better type matching, and more descriptive typing missing?
> 
> I think I might just be lacking clarity on the benefit of shifting from a well-known implementation of `Result` => home-rolled `tuple[Success, Failure]`
> 
> Q2: Why would you have a list of errors? Can a transformation fail in multiple ways? All good id so, then having `[Success, list[Failure]]` makes sense

q1 - I think for me `Result => [Success,Failure]` suggest that there are two paths/two outputs. The way that we have constructed this process ensures that a document can never fail a transformation, it either is fully transformed or is transformed but the associated labels are not transformed alongside it. If we don't want to ever skip/ignore documents in this instance, i can't see how it falls into that `Result` type because what exactly is the failure path.

q2 - A case/document/family may have multiple legal concepts that fail to transform into labels. So not necessarily one document failing transformation, but more one document having multiple labels that have failed to transform. @jamesgorrie

odrakes-cpr:
@jamesgorrie

```
def transform_navigator_family(
    input: Identified[NavigatorFamily],
) -> TransformResult:
    logger = get_logger()
    with log_context(import_id=input.id):
        logger.info(f"Transforming family with {len(input.data.documents)} documents")
        result = transform(input)
        logger.info(
            f"Transform completed: produced {len(result.documents)} documents, "
            f"{len(result.errors)} errors"
        )
        return result
```

Happy update this return type to be [Success(result),Failure], but I am struggling to see the logic when one path is unreachable

jamesgorrie:
@odrakes-cpr nice thanks for clarifying! I think I was missing the "label sub-errors" case.


If we trying to stay in functional FP land, and maintain the `Result` as the core return type which avoids people needing to know when we do or do not and having a constant return type, I think we could just extend your code a little.

```python
class TransformLabelWarning(BaseModel):
    id: str
    message: str

class OtherWarning(BaseModel):
    id: str
    message: str

class TransformError(BaseModel):
    message: str

class Label(BaseModel):
    id: str
    name: str

class Document(BaseModel):
    labels: list[Label]

class TransformResult(BaseModel):
    document: Document
    warnings: list[TransformLabelWarning | OtherWarning]

def transform_label(nav_label) -> Result[Label, TransformLabelWarning]:
    if nav_label.get("name"):
        return Success(Label(id=nav_label["id"], name=nav_label["name"]))
    else:
        return Failure(TransformLabelWarning(id=nav_label["id"], message="missing name"))

def transform_labels(nav_fam) -> list[Result[Label, TransformLabelWarning]]:
    return [transform_label(nav_label) for nav_label in nav_fam.get("labels", [])]

def transform(nav_fam) -> Result[TransformResult, TransformError]:
    label_results = transform_labels(nav_fam)

    labels: list[Label] = []
    warnings: list[TransformLabelWarning | OtherWarning] = []

    for result in label_results:
        match result:
            case Success(label):
                labels.append(label)
            case Failure(warning):
                warnings.append(warning)  # non-fatal, accumulate

    document = Document(labels=labels)
    return Success(TransformResult(document=document, warnings=warnings))


```

data-in-pipeline/app/transform/models.py
```diff
@@ -8,3 +13,47 @@ class NoMatchingTransformations(Exception):
...
+    """
+
+    kind: Literal["unknown_geography"] = "unknown_geography"
+    family_import_id: str
+    geography_id: str
+
+
+# Discriminated union — `kind` is the discriminator field.
+# @see: https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions
+type TransformWarning = Annotated[
```
annaCPR, line 56: Nice!

annaCPR (approved):
lgtm

PR #1222, opened 2026-05-05
Feature/app 1487 transform unfccc gst1

annaCPR:
# What does this do?

Adds a `GST1 Submission` label to all documents that are:

- part of the `UNFCCC.corpus.i00000001.n0000` corpus
- and have been created on the following dates: `2023-11-30` (for party documents), [`2023-11-30`, `2024-10-15`,
`2024-10-17`, `2024-11-06`, `2024-11-15`, `2024-11-18`] (for non-party documents)

Makes the `created` field available on the families and documents served from the `families-api`

## Why is it needed?

To support future work of allowing users to search for GST1 submissions.

## How was it tested?

Unit tests

### DIFF
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -766,9 +766,9 @@ def _transform_litigation_data(
     return Success(labels)
 
 
-def _part_of_gst1(navigator_family: NavigatorFamily) -> bool:
+def _part_of_global_stock_take_1(navigator_family: NavigatorFamily) -> bool:
     """
-    Checks if a family was fart of the first Global Stocktake.
+    Checks if a family was fart of the first Global Stocktake (GST1).
     """
     gst1_party_submission_date = "2023-11-30"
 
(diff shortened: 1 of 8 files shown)


data-in-pipeline/app/transform/navigator_family.py
```diff
@@ -764,6 +766,43 @@ def _transform_litigation_data(
     return Success(labels)
 
 
+def _part_of_global_stock_take_1(navigator_family: NavigatorFamily) -> bool:
```
jamesgorrie: nit: we use gst1 everywhere else - should we use it here? We also seem to have a mix of caps/lowercase.

odrakes-cpr (approved):
i'm guessing we don't have any internal documentation anywhere to support this business logic that we are defining in code.

jamesgorrie (approved):

data-in-pipeline/app/transform/navigator_family.py
annaCPR: I had it as gst1 originally and explained it in the docstring but then I thought it would be nice to be explicit. I can change it back if that's better

annaCPR:
Only in the original ticket @odrakes-cpr

PR #1223, opened 2026-05-06
fix: use search ECS service for search-api

jamesgorrie:
# What does this do?

- uses the export from here as the [new ECS service](https://github.com/climatepolicyradar/search/pull/254)

## Why is it needed?

- [ECS migration](https://linear.app/climate-policy-radar/issue/APP-2035/apprunner-ecs-express-migration)

Depends on https://github.com/climatepolicyradar/search/pull/254

[RFC](https://www.notion.so/climatepolicyradar/RFC-Migrating-services-off-AppRunner-33d9109609a4804c9538e70bbb153101) (🔒)


katybaulch (approved):
lgtm

PR #1224, opened 2026-05-06
Deduplicate collection documents and fix links to families

annaCPR:
# What does this do?

Adds a function to combine duplicate collection document entries with a single link each to a different related family in the transformed response into one collection document which includes all of the related families.

## Why is it needed?

We were only including a link to a single family on collection documents, even if multiple families were part of a collection.

## How was it tested?

Unit tests


odrakes-cpr (approved):
LGTM

PR #1225, opened 2026-05-07
fix: transform entity_type => role

jamesgorrie:
# What does this do?
- swaps `entity_type` => `role`

## Why is it needed?

- mainly because the data is a little messy and we need to differentiate

Towards https://linear.app/climate-policy-radar/issue/APP-2015/removing-document-role-from-entity-type-list


katybaulch:
> # What does this do?
> * swaps `entity_type` => `role`
> 
> ## Why is it needed?
> * mainly because the data is a little messy and we need to differentiate
> 
> Towards https://linear.app/climate-policy-radar/issue/APP-2015/removing-document-role-from-entity-type-list

Will this break anything in the frontend? I vaguely remember seeing `entity_type` labels scattered across the shadow search pages, but not sure if they're hooked up to anything?

jamesgorrie:
@katybaulch nice Q - A: No

> I don't think so, there's not been a need in the transformation work to look at value.type

-- @jesskelsall

PR #1227, opened 2026-05-07
Fix local dev - load-api

annaCPR:
# What does this do?

- Adds a libfaketime dependency for freezing time in the container and multistage builds in the Dockerfile so only the dev container has the dev dependency but share everything else with the prod build
- Removes the old and now unused env vars referring to master DB credentials
- Updates conftest.py to use the real engine instead of creating a new one for tests only (this way tests go through the same code as production, including authentication via token)
- Adds an ability to provide a dump of Aurora data to the dev DB for a full experience ;)
- Adds a README
- Updates test_aws.py to use monkeypatch to manipulate env vars in test rather than directly changing them in the os.environ (this was unsetting env vars for other tests in Docker)

# Why is it needed?
- The implementation of IAM authentication and removal of master credentials from the code meant that local instances of data-in-api could no longer connect to their respective database as dynamically generated auth tokens were being used as the connection password.
- Also, test fixtures have drifted from production implementation and were still using regular password auth to connect to the test db
- boto3 does not call out to AWS to get a token but generates one itself based on the following:
current time
DB_URL
DB_NAME
DB_USERNAME
AWS_REGION
DB_PORT
AWS_ACCESS_KEY_ID:
AWS_SECRET_ACCESS_KEY
AWS_SESSION_TOKEN

Based on that we were able to generate a predictable token to use as the password for both the dev and test DBs by providing the required env vars in the docker containers and freezing time


odrakes-cpr (approved):
lgtm

PR #1230, opened 2026-05-12
fix: add some loose label types

jamesgorrie:
# What does this do?

- adds more granularity to our label types
  - `author`
  - `author_type`
  - `stakeholder_type`
  - `multilateral_climate_fund`
  - `un_convention`

I do not think these are very good types to add to our taxonomy, but they are also not hard to clear out and refine once we have migrated over to using the new search<sup>1️⃣</sup>.

This also follows a pattern of 1 transformer / `label.type` which @annaCPR and I had spoken about - and I think this makes it much more clearer and easy to describe and understand. 👉 https://linear.app/climate-policy-radar/issue/APP-2088/move-transformnavigator-family-1-transformer-label-pattern

## Why is it needed?

- we need this to replicate the legacy search filters on new search

## How was it tested?

Pytest

---

## <sup>1️⃣</sup> Example

UNCCD` or `Adaptation Fund` should probably both be `agent`s, and related to the Document as e.g.

```
Document -- provided_by --> agent::Adaptation Fund
Document -- part_of --> agent::Adaptation Fund

Document -- provided_by --> agent::UNCCD
Document -- submitted_to --> agent::UNCCD
```

Allowing you then query for anything to do with that `Label`. We could then ask for anything to do with that label, or specific questions. This would require us to relatively heavily refactor of search to facet on relationship. The requirements in that project feel to loose to warrant that work though.

👉  [tech debt captured here](https://linear.app/climate-policy-radar/issue/APP-2087/investigate-whether-faceting-on-relationships-makes-more-sense-for-the)

Toward https://linear.app/climate-policy-radar/issue/APP-2084/extend-labels-to-support-current-search-filters

### DIFF
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -31,6 +31,30 @@ from app.transform.models import (
 # Constants
 # ---------------------------------------------------------------------------
 
+_corpus_to_provider_map = {
+    "CCLW.corpus.i00000001.n0000": "Grantham Research Institute",
+    "Academic.corpus.Litigation.n0000": "Sabin Center for Climate Change Law",
+    "CPR.corpus.Goldstandard.n0000": "Gold Standard",
+    "CPR.corpus.i00000589.n0000": "Naturebase",
+    "CPR.corpus.i00000001.n0000": "NewClimate Institute",
+    "CPR.corpus.i00000002.n0000": "Climate Policy Radar",
+    "CPR.corpus.i00000591.n0000": "Laws Africa",
+    "CPR.corpus.i00000592.n0000": "UNDRR",
+    "MCF.corpus.AF.n0000": "Adaptation Fund",
+    "MCF.corpus.AF.Guidance": "Adaptation Fund",
+    "MCF.corpus.CIF.n0000": "The Climate Investment Funds",
+    "MCF.corpus.CIF.Guidance": "The Climate Investment Funds",
+    "MCF.corpus.GCF.n0000": "Green Climate Fund",
+    "MCF.corpus.GCF.Guidance": "Green Climate Fund",
+    "MCF.corpus.GEF.n0000": "Global Environment Facility",
+    "MCF.corpus.GEF.Guidance": "Global Environment Facility",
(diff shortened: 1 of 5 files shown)


data-in-pipeline/app/transform/navigator_family.py
```diff
@@ -1436,3 +1446,138 @@ def _deduplicate_labels(
...
+                    value=author,
+                    type="author",
+                ),
+            )
+        )
+    return labels
+
+
+# region author_type label
+_stakeholder_types = ["Party", "Non-Party"]
```
annaCPR, line 1479: Question: Aren't stakeholder_type and author_type the same thing at least at the moment? I'm not aware of author_types other than Party/Non-Party. Or are we looking ahead to future data?

data-in-pipeline/app/transform/navigator_family.py
```diff
@@ -1436,3 +1446,138 @@ def _deduplicate_labels(
...
+
+
+def _author_type_label(
```
annaCPR, line 1482: Note: I think this will break the FE as it relies on author_type and author values coming from the same label. This is the only case where that happens as far as I'm aware so it might make the FE code cleaner. Apologies, I realise you asked about this in Slack a few days ago but I didn't pick up on it at the time.

annaCPR (approved):
I really like this refactor - I think it will not only make our label types tighter and more controlled in the end but also easier to manage and understand in the code.

data-in-pipeline/app/transform/navigator_family.py
jamesgorrie: Nope - we have to be able to differentiate between them. Stakeholder type only includes Party / Non-Party.

`author_type` includes everything else.

data-in-pipeline/app/transform/navigator_family.py
jamesgorrie: @annaCPR is this true given that this is additive and we maintain the current way of representing `author_type::author` further up the file?

annaCPR: Good point - I think you're right.

annaCPR (approved):
lgtm

PR #1232, opened 2026-05-13
Feature/app 2069 bug missing litigation documents

annaCPR:
# What does this do?

- Adds a new `status` value specifically for litigation documents that were transformed from events due to document files not being available.

## Why is it needed?

- FE filters out any documents that do not have `status.published` but it doesn't make sense to set that status on the documents from events, since they would not have been `published`. Hence a new status to explicitly describe the state of those documents and allow the FE to include them.

## How was it tested?
Unit tests


data-in-pipeline/app/extract/connectors.py
```diff
@@ -26,12 +27,25 @@ class NavigatorEvent(BaseModel):
...
+    PUBLISHED = "published"
+    DELETED = "deleted"
+    CREATED = "created"
+
+
+# This is a new status we are introducing to allow documents transformed
+# from litigation events (when there is no file available) to be distinguished
+# from regular published documents.
+class LitigationDocumentStatus(Enum):
+    PENDING_DOCUMENT_FILE = "pending_document_file"
```
odrakes-cpr: How do these sound 

```
PENDING_SOURCE_FILE
AWAITING_SOURCE_FILE
FILE_PENDING
SOURCE_DOCUMENT_PENDING
```

annaCPR: 😆 honestly they all sound equally good to me

annaCPR: going with `AWAITING_SOURCE_FILE`

odrakes-cpr (approved):

PR #1233, opened 2026-05-14
Remove shipyard urls and replace with apprunner for CORS

annaCPR:
# What does this do?

Attempts to resolve CORS issues for ephemeral environments used in the FE CI


katybaulch (approved):
thankyou so much!! hope this works 🤞🏻

PR #1234, opened 2026-05-14
Update CCC data download README

katybaulch:
# What does this do?

What it says on the tin

## Why is it needed?

Dominyka asked me to!

## How was it tested?

n/a

Note that this code is spaghetti and very soon we will be generating these dumps from the data lake, so i'm not fussed on spending longer on a solution to this!


annaCPR (approved):
lgtm

PR #1237, opened 2026-05-18
Feature/app 2065 map domain to data in model label

odrakes-cpr:
# What does this do?

Map new metadata types to labels. 

## Why is it needed?

This has been mapped in our RDS database, we need to ensure this flows through for data consistency. 

## How was it tested?
Updated transform test suite


annaCPR (approved):
lgtm

PR #1243, opened 2026-05-20
Fix no transformation warnings should be sent for custom id lookups

odrakes-cpr:
per title we were getting lots of transformation errors for unknown geographies when trying to read the regions for XAB and EUR, as these are custom ids they will not have world bank region. 

As such I am skipping the transformation warning as this is not applicable in this sense. 

These were the errors coming through in the AWS logs. 

```
"kind='unknown_geography' family_import_id='CCLW.family.i00006367.n0000' geography_id='EUR'",
"kind='unknown_geography' family_import_id='CCLW.family.i00006367.n0000' geography_id='EUR'",
"kind='unknown_geography' family_import_id='Sabin.family.92290.0' geography_id='XAB'",
"kind='unknown_geography' family_import_id='Sabin.family.92290.0' geography_id='XAB'",
```

### DIFF
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -40,7 +40,7 @@ from app.transform.models import (
 # Constants
 # ---------------------------------------------------------------------------
 
-_custom_countries_by_id = {c.id: c for c in custom_countries}
+_eu_and_international_region_ids = {c.id: c for c in custom_countries}
 
 _corpus_to_provider_map = {
     "CCLW.corpus.i00000001.n0000": "Grantham Research Institute",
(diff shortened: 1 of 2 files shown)


data-in-pipeline/app/transform/navigator_family.py
```diff
@@ -624,7 +627,7 @@ def _transform_geographies(
                         value=region_label,
                     )
                 )
-            else:
+            elif geography.id not in _custom_countries_by_id:
```
jamesgorrie: As an alternative - could we make EU + International Regions?

katybaulch: We used to have 'no region' for these

odrakes-cpr: hmm I don't have a preference, so will go for the first suggestion.

annaCPR (approved):
nice catch

PR #1244, opened 2026-05-21
fix: use label type transformers and add subconcept_of

jamesgorrie:
# What does this do?

- moves `category`, `deprecated_category` and partial `entity_type` to the clearer label transformers
- adds `subconcept_of` to MCFs for `Project` & `Guidance` labels 
- removes `Project` & `Guidance` from non-Principal documents (this was a mistake)
- adds `subconcept_of` to `author` for `Corporate Disclosure`
- adds more granular label type tests

## Why is it needed?

- to start 

## How was it tested?

- pytest

The tests are mostly reordering of the labels

---

part of https://linear.app/climate-policy-radar/issue/APP-2084/extend-labels-to-support-current-search-filters

part of https://linear.app/climate-policy-radar/issue/APP-2088/move-transformnavigator-family-1-transformer-label-pattern

PR #1246, opened 2026-05-21
fix: add law and MCF label types

jamesgorrie:
# What does this do?

- adds labels for laws and MCFs

Follows [this pattern](https://github.com/climatepolicyradar/navigator-backend/pull/1244).

None of this should break frontend as it is all additive, or no-ops.


katybaulch (approved):
nice adds. can see how this is going to get complicated!

PR #1250, opened 2026-05-26
Feature: Pla 1383 - Backend Api Exact Match Default to True

THOR300:
# What does this do?
Configures the backend api to running `exact_match=True` searches as default. This is as we've taken the decision to deprecate embeddings in the backend as outlined in this [notion](https://www.notion.so/climatepolicyradar/RFC-Embeddings-Deprecation-from-Platform-3599109609a4801fad2cc8caef57a09d?d=4dd9109609a4834988a6835c7f57e605) document. The frontend already defaults to `exact_match=True`, this is an additional layer of protection. 

The `mutate_search_body_for_search_type` function can override this to setting exact match as `False`, however this is fine because the transform also sets `all_results=True` which results in a non-embeddings based search against vespa. 

We're not doing a wholesale removal of embeddings from the backend api and query layer (including the cpr_sdk) because we're about to migrate to the search / data-in api and schema. 

## Why is it needed?
Embeddings are being removed from vespa, the pipeline will no longer produce them, thus we need to negate the ability to query on that path. 

## How was it tested?
- Existing unit tests.


backend-api/app/models/search.py
```diff
@@ -67,6 +67,18 @@ class SearchRequestBody(CprSdkSearchParameters):
     For example: 'Adaptation strategy'"
     """
 
+    exact_match: bool = Field(
+        default=True,
+        description=(
+            "Whether to only return results that exactly match the query string. "
+            "This defaults to True as semantic search is deprecated in the backend."
+        ),
+    )
```
jamesgorrie, line 76: note (non-blocking): this still allows an invalid value to be passes (`False`). 

We could force a `True` value
```python
class SearchRequestBody(BaseModel):
    exact_match: bool = True

    @field_validator("exact_match", mode="before")
    @classmethod
    def force_exact_match(self, v):
        return True  # silently ignore whatever was passed
```

This would avoid breaking any services that do send an invalid value, or, if we wanted we could actually validate using

```python
exact_match: Literal[True] = True
```

Just some idle thoughts.

jamesgorrie (approved):
noice

backend-api/app/models/search.py
THOR300: Yes! Great suggestions, I was thinking about something along these lines. After starting I realised it's a little more complex than was initially thought!

We can't set the field to `Literal[True]` because it is still actually mutated to being `False` in some instances: 
- https://github.com/climatepolicyradar/navigator-backend/blob/main/backend-api/app/service/search.py#L478

The `force_exact_match` solution then doesn't protect us against something like the above mutate function where we override the attribute that's after instantiation of the object but as you say it does protect us from an initial `False` value. 

I put a write up of why we're still fine to remove embeddings here: 
-  https://linear.app/climate-policy-radar/issue/PLA-1384/deprecate-embeddings-from-vespa#comment-620f6c8f

THOR300: Going to look at implementing the field validator method 👍

THOR300: This caused multiple test failures so going to stick with the original implementation.

PR #1251, opened 2026-06-01
Updating the deployment to utilise the prefect Variables.

THOR300:
# What does this do?
- Migrated from the ECSVariables custom prefect block type to Variable for prefect config (job variables).
- Updated the prefect dependency version to the latest release now that we're not blocked.

## Why is it needed?
- We temporarily integrated the ECSVariables as the JSON type we originally used was being deprecated, now we're doing the upgrade so can stop using the ECSVariables and instead use the Variable block type. 

## How was it tested?
- Existing CI checks.


odrakes-cpr (approved):
nice, had a mental note to go in and change this myself.

PR #1252, opened 2026-06-01
Bugfix: Correct production aws env to prod for prefect variable load 

THOR300:
# What does this do?
Fixes a failing CI deploy due to the disparity between the aws_env used in this module and in the `Variable` name. 

The variable name in prefect: 
image

## Why is it needed?
CI was failing. 

## How was it tested?
Existing unit tests.


odrakes-cpr (approved):
lgtm

PR #1259, opened 2026-06-04
feat: add ecs service to concepts api

odrakes-cpr:
## Add ECS Express Gateway service to concepts API

Mirrors the ECS setup introduced for the geographies API, giving the concepts API a second compute backend alongside the existing App Runner service.

### Changes
- Added `ExpressGatewayService` resource using the shared cluster, task execution role, and infrastructure role pulled from the `ecs-infra` stack reference
- Added stack references to `aws_env` for subnet IDs across all three AZs
- Created a dedicated ECS task role (`concepts-api-{stack}-ecs-task-role`) with SSM access scoped to `concepts-api/*`
- Exported `ecs_express_service_url` alongside the existing `apprunner_service_url`

### Notes
- All existing App Runner resources are unchanged
- No S3 policy attached to the task role — concepts API has no bucket dependency (unlike geographies)
- CPU/memory set to `1024`/`2048` to match geographies defaults

## Why is it needed?
[Ongoing ECS Migration ](https://www.notion.so/climatepolicyradar/Service-Migration-from-AppRunner-to-ECS-Express-SMD-3599109609a4801a8546daa6cc41a766?source=copy_link)
## How was it tested?
Testing that the staging urls are up and we can see the openapi spec

### DIFF
--- a/concepts-api/infra/__main__.py
+++ b/concepts-api/infra/__main__.py
@@ -1,7 +1,18 @@
 import pulumi
 import pulumi_aws as aws
+from pulumi_aws.ecs.express_gateway_service import (
+    ExpressGatewayService,
+    ExpressGatewayServiceNetworkConfigurationArgs,
+    ExpressGatewayServicePrimaryContainerArgs,
+    ExpressGatewayServicePrimaryContainerEnvironmentArgs,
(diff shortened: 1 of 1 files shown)


annaCPR (approved):
lgtm

PR #1260, opened 2026-06-04
fix: add health endpoint to app root for ALB health checks

odrakes-cpr:
# What does this do?

our ECS container won't start up because it can't reach the health endpoint, the actual route defintion is /concepts/health but as you can see by Amazon Q's diagnosis this causes an issue if you set that endpoint as the health check

```
Based on my comprehensive investigation, I can now provide you with a complete analysis of the ECS service deployment issue.
ROOT CAUSE ANALYSIS
Primary Root Cause: Health Check Endpoint Mismatch
The deployment is failing due to inconsistent health check endpoint configurations between the two target groups:

Target Group 1 (ecs-gateway-tg-9ad4e83c5e4f66ac6): Health check path = /concepts/health
Target Group 2 (ecs-gateway-tg-bb115eb70d0af98fc): Health check path =
/health
Current State Analysis:
Task Status:

Task 1 (af82e5a8ec55...): Running task definition revision 2, IP 10.0.20.109
Container status: RUNNING
Health status: UNKNOWN (no container-level health checks configured)
Registered to target group expecting /concepts/health endpoint
Target group reports: UNHEALTHY - 404 errors
Task 2 (60e7eb7c18c2...): Running task definition revision 1, IP 10.0.15.220
 
Container status: RUNNING
Health status: UNKNOWN
No targets registered in its target group Secondary Issues:

Rollback Failure: The "ROLLBACK_FAILED" status with "No rollback candidate was found" indicates that both deployments have failed, leaving no healthy deployment to roll back to.
Circuit Breaker Activation: Both deployments show "ECS deployment circuit breaker: tasks failed to start" due to repeated health check failures.
Application Health Check Implementation: The 404 responses suggest the application either:
Doesn't implement the /concepts/health endpoint
Has a routing issue preventing access to the health endpoint
Is not fully initialized when health checks begin RESOLUTION RECOMMENDATIONS Immediate Actions:

```

so i have added an app level health endpoint, which replicates the concept/health

### DIFF
--- a/concepts-api/app/main.py
+++ b/concepts-api/app/main.py
@@ -303,7 +303,7 @@ async def get_concept(concept_id: str):
     return {"concept": concept, "related_concepts": related, "subconcepts": subconcepts}
 
 
-@router.get("/health")
+@router.get("/db-health")
 async def health_check():
     db = get_db()
 
--- a/concepts-api/app/main.py
+++ b/concepts-api/app/main.py
@@ -89,17 +89,10 @@ app = FastAPI(
 
 @app.get("/health")
 async def service_health_check():
-    db = get_db()
-    try:
-        db.execute("SELECT 1").fetchone()
-        return {
-            "status": "ok",
-            "version": settings.github_sha,
-        }
-    except Exception:
-        raise HTTPException(
-            status_code=500, detail="Database connection failed"
-        ) from Exception
+    return {
+        "status": "ok",
+        "version": settings.github_sha,
+    }
 
 
 @router.get("/search")
--- a/concepts-api/app/main.py
+++ b/concepts-api/app/main.py
@@ -136,9 +136,10 @@ async def search_concepts(
                 """,
                 [limit],
             )
-    elif has_classifier:
-        result = db.execute(
-            """
+    else:
+        if has_classifier:
+            result = db.execute(
+                """
                 SELECT
                     wikibase_id,
                     preferred_label,
@@ -153,11 +154,11 @@ async def search_concepts(
                 AND has_classifier = ?
                 LIMIT ?
                 """,
-            [f"{q}%", has_classifier, limit],
-        )
-    else:
-        result = db.execute(
-            """
+                [f"{q}%", has_classifier, limit],
+            )
+        else:
+            result = db.execute(
+                """
                 SELECT
                     wikibase_id,
                     preferred_label,
@@ -171,8 +172,8 @@ async def search_concepts(
                 WHERE preferred_label ILIKE ?
                 LIMIT ?
                 """,
-            [f"{q}%", limit],
-        )
+                [f"{q}%", limit],
+            )
 
     if result.description is not None and (rows := result.fetchall()):
         columns = [desc[0] for desc in result.description]
--- a/concepts-api/app/main.py
+++ b/concepts-api/app/main.py
@@ -27,7 +27,7 @@ try:
         ServiceManifest.from_file(f"{root_dir}/service-manifest.json"), ENV, "0.1.0"
     )
 except Exception as _:
-    _LOGGER.error("Failed to load service manifest, using defaults")
+    _LOGGER.exception("Failed to load service manifest, using defaults")
     otel_config = TelemetryConfig(
         service_name="navigator-backend",
         namespace_name="navigator",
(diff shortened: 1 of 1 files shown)


concepts-api/app/main.py
```diff
@@ -88,6 +87,21 @@ class Settings(BaseSettings):
 )
 
 
+@app.get("/health")
+async def service_health_check():
+    db = get_db()
+    try:
+        db.execute("SELECT 1").fetchone()
```
henry-cpr: OK, so I know this is the way we've been doing healthchecks forever here, but I think we need to consider it again with the move to ECS. 

when a task fails a health check its not just a container that gets taken out of rotation, the task is fully replaced, involving stopping it and launching a new container -- which is an expensive operation (ie its a restart decision rather than a load-balancer routing decision for this). 

In general with healthchecks you want to separate out whether you're asking "is this task responding" versus "is this task ready to receive traffic". the latter feels like a good place to include a DB check, because you only want to receive traffic if the DB is responding, but ECS treats healthchecks as the former. A DB outage will likely affect all instances equally, so healthchecks will fail, and ECS will start  relaunching all the containers, which is not what we want it to do when the DB fails; it wont fix anything! More than that, it might create cascading failures as mass container restarts cause more healthchecks to surge and overwhelm the db all over again.  

I would move the DB check to a separate DB health endpoint, and we'll get grafana to consume and alert on that later at a reduced frequency. We have DB alerts so im happy DB availability in general is covered, but "ECS can't talk to DB" is slightly distinct from that and not covered. But I think i'd rather have that gap than risk ECS container churn and DB connection storms...

concepts-api/app/main.py
```diff
@@ -88,6 +87,21 @@ class Settings(BaseSettings):
...
+        return {
+            "status": "ok",
+            "version": settings.github_sha,
+        }
+    except Exception:
+        raise HTTPException(
+            status_code=500, detail="Database connection failed"
+        ) from Exception
```
henry-cpr: I think this needs to be except Exception as e: ... raise ... from e; youre raising from the Exception class not the exception instance that has been raised

concepts-api/app/main.py
```diff
@@ -147,11 +160,11 @@ async def search_concepts(
                 AND has_classifier = ?
                 LIMIT ?
                 """,
-                [f"{q}%", has_classifier, limit],
-            )
-        else:
-            result = db.execute(
-                """
+            [f"{q}%", has_classifier, limit],
```
henry-cpr: these changes aren't related to the healthcheck, sholuld they be in here?

concepts-api/app/main.py
henry-cpr: It's occurred to me that this is a duckdb service so i'm probably overblowing things, I had in my head that this was talking to RDS.  So less of a thing. But still worth considering.

henry-cpr (changes requested):
A couple of comments that need attending to please. I must admit i'm not convinced by the Q analysis either. In my experience this sort of issue is much more likely a infra config/ALB issue -- have you ruled that out? I would prefer to have AWS consistently pinging health endpoints than duplicating health endpoints in our code, and I don't fully understand from the context I've look at here so far why the target groups are having the issue. If the /concept/health endpoint is defined in the code, then surely the 404 means the target group is misconfigured if it cant find it? Like the URL it's trying to ping is wrong?

concepts-api/app/main.py
odrakes-cpr: trunk fmt sadly

concepts-api/app/main.py
odrakes-cpr: Don't want this discussion lost, it should have happened here - but I responded to [your comment on slack ](https://climate-policy-radar.slack.com/archives/C0B264BE9GE/p1780568463388099)for visibility. 

TLDR 

We're deploying the concepts API to ECS using a shared cluster/ALB setup (the same infrastructure that the geographies API runs on successfully). The deployment is stuck — the ECS circuit breaker is firing because the ALB health checks are returning 404s.

The ALB is configured to ping /health (consistent with the shared infra setup), but in main.py the health endpoint is registered under the /concepts router prefix, meaning it only exists at /concepts/health. The logs confirm the request is reaching the container:
INFO 10.0.41.83:18778 - "GET /health HTTP/1.1" 404

The geographies API (which is healthy and running on the same cluster) has its health endpoint registered at the app root level, not under a prefixed router.

henry-cpr (approved):
looking  much nicer!

PR #1261, opened 2026-06-04
Feature/pla 1384 deprecate embeddings from backend api 

THOR300:
# What does this do?
- Wholesale removal of the backend-api's ability to query the embeddings fields in vespa. This is done here by integrating a bump to the cpr sdk's package to a recent release that clears embeddings from the query layer: https://github.com/climatepolicyradar/cpr-sdk/releases/tag/v5.0.0

- We're clearing embeddings down from the query layer (backend-api and cpr_sdk) before we then go and remove the field and ranking profiles from vespa itself. 

- Some updates to the tests following the query changes, explanations commented against the modules themselves. 

## Why is it needed?
- Vespa costs us $13k / month, removing embeddings we estimate halves that cost. 
- We're not using embeddings so it's good practice to clear down old code paths etc. from our tech stack. 

## How was it tested?
- Existing unit and integration tests.


backend-api/tests/search/vespa/fixtures/vespa_document_passage.json
THOR300, line 1: note: Added passages containing "climate" to CCLW.executive.10246.4861 (2 passages) and CCLW.executive.4934.1571 (2 passages). cpr-sdk 5.0.0 replaced nearestNeighbor with pure keyword search, so test_continuation_token__passages — which queries for "climate" — now requires passages that literally contain the word.

backend-api/tests/search/vespa/setup_search_tests.py
THOR300, line 1: note: _populate_db_families previously deduped at the family level but not the document level. Adding a second passage for CCLW.executive.10246.4861 caused a primary key violation on FamilyDocument since _create_document was called twice for the same document_import_id. Added seen_document_ids tracking to guard against this.

backend-api/tests/search/vespa/test_vespa_search_pagination.py
THOR300, line 1: note: With keyword-only search each target family has 2 matching passages. At the default max_hits_per_family=10 Vespa returns both at once, so no passage continuation token is generated. Setting it to 1 in the test ensures Vespa pages through passages one at a time, which is what the test is exercising.

odrakes-cpr (approved):

jamesgorrie (approved):
Show me the 💰 ! Nice work.

PR #1262, opened 2026-06-04
Feature/pla 1384 bumping the cpr_sdk version to latest.

THOR300:
# What does this do?
- Updates the cpr sdk version for the backend api to the latest version. 
- This included updating post a couple of breaking changes where some fields were renamed. We were returning the response model of the sdk in the `families` and `documents` endpoints, this would've allowed the breaking change to propagate through so I've added an interrim model to stop this happening. 
- We can see the diff between the two different versions at this [link](https://github.com/climatepolicyradar/cpr-sdk/compare/v1.23.4...v4.3.0). This mainly consists of the addition of `PassagesV2`, `ConceptsV2` as well as some refactoring work. 👈 Please can the reviewer from @climatepolicyradar/application_engineers take a quick look at this as well 🙏 

## Why is it needed?
- It was out of date and we need to bring in a new release to facilitate the embeddings deprecation. 

## How was it tested?
- Existing unit and integration tests.


backend-api/app/models/search.py
```diff
@@ -373,6 +375,37 @@ def increment_pages(self):
...
+        return cls(
+            total_hits=r.total_hits,
+            total_family_hits=r.total_result_hits,
+            families=r.results,
+            query_time_ms=r.query_time_ms,
+            total_time_ms=r.total_time_ms,
+            continuation_token=r.continuation_token,
+            this_continuation_token=r.this_continuation_token,
+            prev_continuation_token=r.prev_continuation_token,
+        )
```
THOR300, line 410: **note**: The `total_result_hits` and `results` fields are the breaking changes here that needed mapping back to the fields which the frontend expects.

backend-api/tests/search/test_search.py
THOR300, line 1: **note**: In these tests the vespa response changes inline with the sdk but the backend's response model stays constant.

.trunk/pyrightconfig.backend-api.json
THOR300, line 1: note: The `pull_request` workflow runs trunk check after installing dependencies from the top level lock files. This doesn't include the backend-api, because we added new imports in `backend-api/app/models/search.py` this caused the check to fail so adding a trunk rule here.

odrakes-cpr (approved):
nice 1

PR #1266, opened 2026-06-09
Feature/app 2176 trigger ecs deployment on image push for geographies api

odrakes-cpr:
## What
Adds a dedicated `deploy-ecs-service.yml` reusable workflow and updates the merge-to-main workflow to trigger ECS redeployments for the four services running on ECS Express (`geographies-api`, `families-api`, `concepts-api`, `data-in-api`).

## Why
App Runner had `auto_deploy: true` which automatically redeployed the service when a new image was pushed to ECR. ECS does not have this behaviour — pushing a new image has no effect on running tasks without an explicit redeployment trigger.

## How
- Adds `ecs_services` to the `changes` job outputs, filtering the changed services list down to only the four ECS Express services
- Adds a `deploy-ecs` job to merge-to-main that calls the new `deploy_ecs.yml` reusable workflow when any of the four services have changed
- The new workflow builds and pushes the image tagged with `${{ github.sha }}` (as well as `latest`), then uses the `amazon-ecs-deploy-express-service` action to deploy the specific SHA-tagged image to ECS
- A verify step re-queries the ECS deployment status after the action completes, since the action can silently return success on timeout or rollback

## Notes
- The original `deploy` job is untouched — `data-in-pipeline`, `data-in-pipeline-load-api` and any future non-ECS services continue to use it
- `concepts-api` service name needs to be renamed to `concepts-api-{environment}` in Pulumi before this workflow will work for that service — that is a follow-up task
- The `navigator-backend-github-actions-deploy` role will need `ecs:ListServiceDeployments` and `ecs:DescribeServiceDeployments` added for the verify step to work


odrakes-cpr:
Screenshot 2026-06-09 at 12 30 07
Screenshot 2026-06-09 at 12 30 16

geographies-api/justfile
```diff
@@ -25,3 +25,18 @@ initial-data environment:
...
+GITHUB_SHA := `git rev-parse HEAD`
+
+# Deployment
+deploy-override service environment tag:
+    just --justfile justfile build {{service}} {{environment}} {{tag}}
+    docker tag {{service}}:{{tag}} $(aws sts get-caller-identity --query 'Account' --output text).dkr.ecr.eu-west-1.amazonaws.com/{{service}}:{{tag}}
+    docker tag {{service}}:{{GITHUB_SHA}} $(aws sts get-caller-identity --query 'Account' --output text).dkr.ecr.eu-west-1.amazonaws.com/{{service}}:{{GITHUB_SHA}}
+    docker push $(aws sts get-caller-identity --query 'Account' --output text).dkr.ecr.eu-west-1.amazonaws.com/{{service}}:{{tag}}
+    docker push $(aws sts get-caller-identity --query 'Account' --output text).dkr.ecr.eu-west-1.amazonaws.com/{{service}}:{{GITHUB_SHA}}
+    aws ecs update-service \
```
odrakes-cpr: This is the bit added on, everything else copied from root

jamesgorrie: There is a [ECS deploy action](https://github.com/aws-actions/amazon-ecs-deploy-express-service) we could potentially start using? [We currently use it on search](https://github.com/climatepolicyradar/search/blob/main/.github/workflows/deploy_api.yml#L39-L49).

geographies-api/justfile
```diff
@@ -25,3 +25,18 @@ initial-data environment:
 dev:
     uv sync
     docker compose --profile dev up
+
+GITHUB_SHA := `git rev-parse HEAD`
+
+# Deployment
+deploy-to-ecs:
```
jamesgorrie: suggestion: remove this

jamesgorrie (approved):

.github/workflows/merge-to-main.yml
```diff
@@ -208,3 +211,13 @@ jobs:
...
+
+  # @see: https://github.com/marketplace/actions/amazon-ecs-deploy-express-service-action-for-github-actions#complete-example-with-all-options
+  deploy-ecs:
+    uses: ./.github/workflows/deploy-ecs-service.yml
+    secrets: inherit
+    needs: [changes, test]
+    if: (needs.test.result == 'success' || needs.test.result == 'skipped') && needs.changes.result == 'success' && needs.changes.outputs.ecs_services != '[]'
+    with:
+      ecs_services: ${{ needs.changes.outputs.ecs_services }}
+      environment: '["staging", "production"]'
```
jamesgorrie, line 223: off topic: I am still really unsure of why / when we deploy / use to staging?

odrakes-cpr: Not sure really\

PR #1268, opened 2026-06-09
fix: remove plural from Corporate Disclosure

jamesgorrie:
# What does this do?

removes the `s` from `Corporate Disclosure`

There is a longer term fix here where we shouldn't be using free text - but it's a larger refactor that I cannot prioritise right now.


katybaulch (approved):
nice interim fix :)

PR #1270, opened 2026-06-09
ensure all microservices follow all the same pattern

odrakes-cpr:
it allows the github workflow for deploys to work seamlessly


annaCPR (approved):
nice

PR #1271, opened 2026-06-09
Feature/app 2176 trigger ecs deployment on image push for geographies api

odrakes-cpr:
the output required for the deploy to ecs to run was not working, so have to update


annaCPR (approved):
lgtm

PR #1272, opened 2026-06-09
fix: add subconcept of report to relevant authors

jamesgorrie:
- adds relevant authors as subconcepts of reports

---


katybaulch (approved):
lgtm

PR #1273, opened 2026-06-09
fix: use litigation for root subconcept of litigation concepts

jamesgorrie:
- uses `category::Litigation` for concepts from Sabin that have no `subconcept_of` from the source data

---


katybaulch (commented):
lgtm!

PR #1279, opened 2026-06-11
update github deploy role 

odrakes-cpr:
Once again I am nearly defeated by aws policies, roles and their allowed actions. 

This gives the deploy role all it requires.


annaCPR (approved):
lgtm

PR #1281, opened 2026-06-15
chore: bump cpu and memory for families api ecs task

odrakes-cpr:
# What does this do?
per title

We tend to hit the ceiling of the task/container resources every day when we run the DIP pipeline.


annaCPR (approved):
lgtm

PR #1282, opened 2026-06-15
fix: add kosovo

jamesgorrie:
# What does this do?

As Kosovo isn't in the ISO-3166 standard, [it has a commonly used user assigned value](https://en.wikipedia.org/wiki/XK_(user_assigned_code)).

`XKX` is used in our DB - [which you can see applied to this family](https://api.climatepolicyradar.org/families/CCLW.family.i00000218.n0000). 

This adds it to our list of custom countries.


odrakes-cpr (approved):

annaCPR (approved):
lgtm

PR #1283, opened 2026-06-15
fix: do not cache partial runs of the data-in-pipeline

jamesgorrie:
# What does this do?

If we run a a partial update (pass in specific IDs to run against) we do not cache the result as it would not be the correct representation and complete output.

We are fine with there potentially being slight drift if the test documents had change as this is a daily run, so there will be eventual consistency with the Aurora data and the S3 cache.


annaCPR (approved):
lgtm

odrakes-cpr (approved):

PR #1285, opened 2026-06-15
Feature/app 2183 trigger ecs deploy on merge to main data-in-pipeline-load-api

odrakes-cpr:
# What does this do?
adds a fix to the previous change made where we tried to deploy `data-in-pipeline-load-api` to ecs but in the shares api cluster, which meant that we ended up creating a dead service. 

This is because `data-in-pipeline-load-api` has its own service and cluster so have updated the flow to reflect that. 

## Why is it needed?

Part of the ECS Migration. 

## How was it tested?

NA


annaCPR (approved):
lgtm

PR #1287, opened 2026-06-16
Rename services to differentiate the new ecs service metrics source

annaCPR:
# What does this do?

As above, adds `-ecs` to the end of each of the service names (`families-api`, `geographies-api`, `concepts-api`) in the service-manifest.json for the services that have been migrated to ECS Express and have OpenTelemetry enabled.

## Why is it needed?

To differentiate which services the metrics are coming from as we migrate from AppRunner to ECS Express and the new and old services run in parallel


odrakes-cpr (approved):
lgtm

PR #1288, opened 2026-06-18
chore: Increase min task count to 2 across ECS express services

odrakes-cpr:
## What
Increases the minimum task count from 1 to 2 on ECS express services.

## Why
A single container getting saturated by request spikes (e.g. from data-in hitting families-api) causes ALB health checks to fail, ECS to kill the task, and a latency gap while a replacement spins up. Running a minimum of 2 containers ensures the ALB can load balance across tasks and maintain availability while ECS replaces any unhealthy container.


annaCPR (approved):
thanks!

PR #1290, opened 2026-06-22
Feature/app 2167 Add missing taiwan region

katybaulch:
Taiwan was missing a region. Now it's not!


jesskelsall (approved):
Easy as that. Approved!

PR #1291, opened 2026-06-22
Allow multiple relationships for same label

annaCPR:
# What does this do?

- Refactor and rename `_transform_family_corpus_organisation` to return a single agent label per given family
- Make agent labels `subconcept_of` Multilateral Climate Fund category label when family is MCF
- Refactor to reuse an existing mapping for MCF corpora IDs to fund names

## Why is it needed?

- So that we can reuse the agent label for different label relationships without accidentally overriding provider labels
- To fix incorrect and also inconsistent fund names in the new `multilateral_climate_fund` label


jamesgorrie (approved):
fabulous.

PR #1297, opened 2026-06-30
Feature/app 2240 test 5xx grafana dashboard and alert

annaCPR:
# What does this do?

Moves the temporary 5xx Testing endpoint to the geographies API

## Why is it needed?

We need to test our 5xx Error Rate Grafana dashboard and Slack alert so using this endpoint to generate some errors. Initially added to the data-in-api but that is currently unavailable on ECS.

### DIFF
--- a/data-in-api/app/main.py
+++ b/data-in-api/app/main.py
@@ -182,13 +182,4 @@ def read_label(label_id: str, db=Depends(get_db)):
         )
 
 
-@app.get("/test-5xx")
-@router.get("/test-5xx")
-def test_5xx():
-    raise HTTPException(
-        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-        detail="Testing 5xx errors for Grafana",
-    )
-
-
 app.include_router(router)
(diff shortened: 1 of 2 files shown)


data-in-api/app/main.py
```diff
@@ -182,13 +182,4 @@ def read_label(label_id: str, db=Depends(get_db)):
...
-@app.get("/test-5xx")
-@router.get("/test-5xx")
-def test_5xx():
-    raise HTTPException(
-        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-        detail="Testing 5xx errors for Grafana",
-    )
-
-
 app.include_router(router)
```
odrakes-cpr, line 185: just a note this will trigger the deploy for the ecs service in production and it will fail because that service despite being deleted is in a weird clean up state

odrakes-cpr (commented):
I will approve, but I would keep the data-in-api changes for now
 or remove it and accept the ci failure.

odrakes-cpr (approved):

PR #1298, opened 2026-06-30
fix: wait for service_healthy on data-in-load-api docker-compose

jamesgorrie:
## Fix flaky test container startup

Two issues were causing just test data-in-pipeline-load-api to fail:

- Race condition: the test container was starting before the DB was ready to accept connections, causing DNS/connection errors. Fixed by adding condition: service_healthy to depends_on so Docker waits for the Postgres healthcheck to pass first.
- Port conflict: the test service was trying to bind port 8080 on the host, which fails if anything else is using it (e.g. Docker Desktop). Pytest doesn't need an exposed port, so the mapping is removed.


annaCPR (approved):
lgtm

PR #1300, opened 2026-06-30
fix: add relationship.type to primary key when storing document => label relationships

jamesgorrie:
# What does this do?

Adds `relationship.type` to the primary key of `Document` => `Label` relationships in the DB.

This allows us to store a `Label` twice against Document, but with a different relationship 

e.g. 
```
Document -- provider --> Global Environment Facility
Document -- part_of --> Global Environment Facility
``` 

Supports https://github.com/climatepolicyradar/navigator-backend/pull/1291


annaCPR (approved):
lgtm

PR #1301, opened 2026-07-01
fix: trailing slash causing errors when hitting related endpoints

odrakes-cpr:
# What does this do?
per title - removes trailing slash.
## Why is it needed?
Error when hitting load api endpoints..

```
Task run failed with exception: HTTPError('500 Server Error: Internal Server Error for url: https://da-9f758dc9847148f1b8df15150d4488fa.ecs.eu-west-1.on.aws//load/run-migrations')
Traceback (most recent call last):
```


annaCPR (approved):
lgtm

PR #1304, opened 2026-07-22
Feature/enri 1518 remove the geographies api app runner service

odrakes-cpr:
# What does this do?
Per title removes the geographies app runner service.

This one has a few stages to it. 

First you unprotect the app runner resource

Then you delete the resource, associated roles and policies

Then you delete the cloudfront distribution from the api stack. 
## Why is it needed?
Clean up work from the ecs migration.
## How was it tested?
Checking that everything is all good in staging.

### DIFF
--- a/api/infra/__main__.py
+++ b/api/infra/__main__.py
@@ -128,16 +128,6 @@ api_cloudfront_distribution = aws.cloudfront.Distribution(
             ),
         ),
         aws.cloudfront.DistributionOriginArgs(
-            domain_name=geographies_api_stack.get_output("apprunner_service_url"),
-            origin_id="geographies-api-apprunner",
-            custom_origin_config=aws.cloudfront.DistributionOriginCustomOriginConfigArgs(
-                http_port=80,
-                https_port=443,
-                origin_protocol_policy="https-only",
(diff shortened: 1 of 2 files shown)


annaCPR:
Question: Are we sure we want to be deleting those generic IAM policies (ECR, S3 and SSM ones)? I know they're attached to Apprunner specific roles but just want to make sure we're not losing permissions.

odrakes-cpr:
> Question: Are we sure we want to be deleting those generic IAM policies (ECR, S3 and SSM ones)? I know they're attached to Apprunner specific roles but just want to make sure we're not losing permissions.

@annaCPR 

the ECS role has it's own SSM policy so that's fine to go.
the s3 policy is not actually deleted, the attachment to the app runner role is being removed, its attached to the ecr role. 
the ecr policy, is again attached to the app runner api role - I am making the assumption that ecs does not need this to access the ecr repository. 

because it is handled in the ecs infra by this managed_policy_arn - "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy",

annaCPR (approved):
all good to go

PR #1307, opened 2026-07-22
Feature/enri 1523 remove the data in api app runner service

odrakes-cpr:
What does this do?
Removes the app runner service from our eco system.

This is clean up from the ecs migration.

Why is it needed?
Whilst it does not cost much, there is a running cost to just leaving this in our estate.

How was it tested?
Tested on staging.


annaCPR (approved):
lgtm

PR #1309, opened 2026-07-27
Feature/enri 1524 remove the families api app runner service

odrakes-cpr:
# What does this do?
This removes the families api app runner service.

This is part of the clean up after the ecs express migration.
## Why is it needed?

Lingering code and lingering applications cost money.
## How was it tested?
Ran on staging.


annaCPR (approved):
lgtm

PR #1310, opened 2026-07-27
:robot: Update vespa test files

cpr-tech-admin:
# What does this do?
Removes 3 indexes from the local test vespa instance. 

## Why is it needed?
The indexes have been removed from the production vespa instance schema. 

## How was it tested?
Existing unit tests in this repo.


annaCPR (approved):
lgtm

PR #1311, opened 2026-07-29
Feature/enri 1534 remove to data in load api app runner service

odrakes-cpr:
# What does this do?

remove the data-in-load-api service. 
This is part of the ecs migration, tidying up and getting rid of the App runner services. 

## Why is it needed?
Everything is served through the equivalent ecs service. 

## How was it tested?
removed on staging first.


annaCPR (approved):
lgtm

PR #1314, opened 2026-08-06
fix: use taxonomical labels for entity type

jamesgorrie:
# What does this do?

- uses taxonomical labels for `entity_type` over free text
- falls back to previous way of doing this to avoid regressions

We have seen this before where
- we label a document with taxonomical labels (has `subconcept_of` relationships)
- we then label it with a free text label
- **we dedupe** 👈 this means we loose the relationships

## How was it tested?

pytest

---

fixes https://linear.app/climate-policy-radar/issue/APP-2261/corporate-voluntary-report-missing-from-corporate-disclosure

### DIFF
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -1500,6 +1500,42 @@ def _transform_document_urls(navigator_document):
     return items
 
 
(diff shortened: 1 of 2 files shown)


annaCPR (approved):
lgtm

PR #1315, opened 2026-08-06
fix: framework law and topic transformations

jamesgorrie:
# What does this do?
- reduces the way we label "Framework law" and `topics` to
  - If anything is in `framework` => "Framework law"
  - If something in `topic` => `topic::{{ topic }}`

## How was it tested?
pytest

---

fixes https://linear.app/climate-policy-radar/issue/APP-2260/law-type-filters-are-all-labeled-as-framework-law


katybaulch (approved):
yum - yes please!

PR #1316, opened 2026-08-06
fix: remove duplicate application of project and guidance label

jamesgorrie:
# What does this do?

- removes duplicate application of "Project" and "Guidance" label, leaving only the taxonomical labels to be applied, which have the `subconcept_of: MCF`.

we have seen this bug before e.g: https://github.com/climatepolicyradar/navigator-backend/pull/1314:
- we label a document with taxonomical labels (has `subconcept_of` relationships)
- we then label it with a free text label
- we dedupe 👈 this means we loose the relationships

## How was it tested?
pytest

---

fixes https://linear.app/climate-policy-radar/issue/APP-2262/report-and-guidance-are-missing-from-mcf-category

### DIFF
--- a/data-in-pipeline/app/transform/navigator_family.py
+++ b/data-in-pipeline/app/transform/navigator_family.py
@@ -1244,31 +1244,6 @@ def _transform_navigator_family(
         )
     )
 
-    # MCF reports and guidance labels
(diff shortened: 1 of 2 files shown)


katybaulch (approved):
nice ✨

